PR #4, opened 2017-08-22
Feature/cab 2544

maximede:


src/app/routing/config-resolver.service.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Promise<Config> {
+
+    let tenant: string = route.params.tenant;
+
+    if (!tenant) {
+      // Try to find a tenant param in the parent routes
+      while (!tenant && route.parent) {
+        tenant = route.parent.params.tenant;
```
agagne: is it possible that a tenant might never be found?

maximede: yes, and it should redirect the user to `/` if that's the case

src/app/pages/tenant/tenant.component.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+        {provide: ActivatedRoute, useValue: activatedRoute},
+        Title,
+      ],
+      schemas: [NO_ERRORS_SCHEMA]
+    })
+      .compileComponents();
+  }));
+
+
+  describe('when component is created and a correct config is sent', () => {
```
huiaic: This description doesn't quite flow with the specs'. How about just "Component", or add "... then the component" at the end.

src/app/widgets/header/header.component.ts
```diff
@@ -0,0 +1,31 @@
...
+import { User } from '../../user/user';
+
+@Component({
+  selector: 'app-header',
+  templateUrl: './header.component.html',
+  styleUrls: ['./header.component.css']
+})
+export class HeaderComponent implements OnInit {
+  tenant: string;
+  promiseUser: Promise<User>;
```
agagne: probably should rename to just User, rather than encoding that it is enveloped in promise in the name

PR #9, opened 2017-08-25
CAB-2518: add husky precommit and prepush hooks for linting & testing

agagne:


package.json
```diff
@@ -25,6 +28,7 @@
     "@angular/router": "^4.0.0",
     "classlist.js": "^1.1.20150312",
     "core-js": "^2.4.1",
+    "karma-spec-reporter": "^0.0.31",
```
maximede: this should be in the dev dependencies

PR #16, opened 2017-09-05
Feature/cab 2562

agagne:


src/app/widgets/content-view/content-view.component.ts
```diff
@@ -0,0 +1,23 @@
...
+  @Input() inputUrl: string;
+
+  constructor(private sanitizer: DomSanitizer) {
+    // TODO: read more about DomSantitizer
+  }
+
+  ngOnInit() {
+    // '282639'; // TODO: example id
+    console.log('Working url: ', this.inputUrl);
+    this.url = this.sanitizer.bypassSecurityTrustResourceUrl(this.inputUrl);
```
maximede: I don't think this is needed. These url should never contain `javascript` or other weird stuff. It should be safe to sanitize the url

from angular doc : 
>For example, when binding a URL in an <a [href]="someValue"> hyperlink, someValue will be sanitized so that an attacker cannot inject e.g. a javascript: URL that would execute code on the website.

src/environments/environment.travis.ts
```diff
@@ -10,5 +10,10 @@ export const environment = {
     context: '/search',
     method: 'POST',
     authenticationHeader: 'auth'
+  },
+  content_api: {
+    url: 'content-api.dev',
```
maximede: missing `http://`

src/environments/environment.ts
```diff
@@ -10,5 +10,10 @@ export const environment = {
     context: '/search',
     method: 'POST',
     authenticationHeader: 'auth'
+  },
+  content_api: {
+    url: 'content-api.dev',
```
maximede: missing `http://`

PR #19, opened 2017-09-15
Feature/cab 2574

agagne:
CAB-2574: As a user, I can edit the metadata on the view and edit page so that it will be stored correctly and other users can see or edit it

### DIFF
--- a/src/app/model/content-item.ts
+++ b/src/app/model/content-item.ts
@@ -1,4 +1,5 @@
 import { isNullOrUndefined } from 'util';
+import * as _ from 'lodash';
 
 export class ContentItem {
   public id: string;
@@ -9,7 +10,7 @@ export class ContentItem {
     if (!isNullOrUndefined(contentItem)) {
       this.id = contentItem.id;
       this.label = contentItem.label;
-      this.metadata = Object.assign({}, contentItem.metadata); // deep copy metadata
+      this.metadata = _.cloneDeep(contentItem.metadata); // deep copy nested metadata
     }
   }
 }
--- a/src/app/model/content-item.ts
+++ b/src/app/model/content-item.ts
@@ -1,5 +1,15 @@
+import { isNullOrUndefined } from 'util';
+
 export class ContentItem {
   public id: string;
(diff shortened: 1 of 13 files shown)


src/app/model/content-item.ts
```diff
@@ -0,0 +1,15 @@
...
+export class ContentItem {
+  public id: string;
+  public label: string;
+  public metadata: Map<string, any> = new Map();
+
+  constructor(contentItem?: ContentItem) {
+    if (!isNullOrUndefined(contentItem)) {
+      this.id = contentItem.id;
+      this.label = contentItem.label;
+      this.metadata = Object.assign({}, contentItem.metadata); // deep copy metadata
```
maximede: Does this work for facilities items that have nested objects in metadata ?

src/app/pages/edit-page/edit-page.component.ts
```diff
@@ -48,6 +48,13 @@ export class EditPageComponent implements OnInit, OnDestroy {
     });
   }
 
+  onSave(savedItem: ContentItem) {
+    this.contentService
+      .update(savedItem)
+      .takeUntil(this.componentDestroyed)
+      .subscribe(); // TODO: handle response
```
maximede: Not sure if this need to be done now, but this is quite important as some wcc rules could update fields that the user didn't change

src/app/services/content.service.ts
```diff
@@ -15,17 +15,31 @@ export class ContentService {
...
-  public read(itemId: string): Observable<ContentResult> {
+  public read(itemId: string): Observable<ContentItem> {
     const url = this.baseUrl + this.itemPathFragment + itemId;
     const options = this.buildRequestOptions();
 
     console.log('Reading content URL:', url);
     return this.http.get(url, options).map(response => {
       console.log('content response: ' + JSON.stringify(response));
-      const contentResult: ContentResult = response.json();
+      const contentResult: ContentItem = response.json();
```
maximede: should `const contentResult` be `const contentItem` ?

PR #32, opened 2017-10-06
Feature/cab 2591

agagne:


src/app/content/content-view/content-view.component.ts
```diff
@@ -12,35 +13,102 @@ import { Subject } from 'rxjs/Subject';
...
+  private displayDefaultUrl(): void {
+    this.url = this.defaultUrl;
+    this.determineItemType();
+  }
+
+  // TODO: type detection should be improved
+  private determineItemType(): void {
+    if (!isNullOrUndefined(this.item)) {
+      // const webExtension = this.item.metadata['WebExtension'];
+      const mimeType = this.item.metadata['MimeType'];
```
maximede: We should probably use constants for that kind of stuff

agagne: agreed, but this is a not the final implementation. :neckbeard:

PR #34, opened 2017-10-10
CAB-2576: Implemented a custom-text directive

jamesrenfro:


src/main.ts
```diff
@@ -3,9 +3,15 @@ import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
...
 import { environment } from './environments/environment';
+// import {getTranslationProviders} from './app/i18n-providers';
 
 if (environment.production) {
   enableProdMode();
 }
 
 platformBrowserDynamic().bootstrapModule(AppModule);
+
+// getTranslationProviders().then(providers => {
```
maximede: Looks like remaining stuff from the i18n impl. Do we still need it ?

jamesrenfro: Thanks - removing it now.

PR #36, opened 2017-10-10
fetch configuration file from github repo

maximede:


src/app/core/shared/config.service.ts
```diff
@@ -1,145 +1,75 @@
...
+import 'rxjs/add/operator/toPromise';
+import { HttpClient } from '@angular/common/http';
+import { environment } from '../../../environments/environment';
 
 @Injectable()
 export class ConfigService {
-  configs: Map<string, Config>;
+  private configs: Map<string, Config> = new Map();
+  private tenantsConfig = null;
+  private listTenantsUrl = 'https://api.github.com/repos/uw-it-edm/content-services-ui-config/contents/' +
```
agagne: should this should be in the environment variable?

src/app/core/shared/config.service.ts
```diff
@@ -1,145 +1,75 @@
...
-                 },
-                "metadata.DocumentYear": {
-                   "key" : "metadata.DocumentYear",
-                   "label" : "Document Year",
-                   "order": "desc",
-                   "size":10
-                }
-             }
-          }
+        if (this.configs.has(tenantInfo.tenantName)) {
```
agagne: what happens if tenantInfo is undefined, will this crash and burn when you try to get tenantName?

PR #38, opened 2017-10-12
use content-api to load user

maximede:


src/environments/environment.travis.ts
```diff
@@ -7,15 +7,19 @@ export const environment = {
...
   search_api: {
-    url: 'search-api.dev',
+    url: 'http://search-api.dev',
     context: '/search',
     method: 'POST',
     authenticationHeader: 'auth'
   },
   content_api: {
     url: 'http://content-api.dev',
     context: '/content/v3',
```
agagne: should we replace context with contextV3?

maximede: that makes sense, I pushed an update for that

PR #39, opened 2017-10-20
Feature/cab 2614

agagne:
we may need to make UI changes

### DIFF
--- a/src/app/content/shared/content.service.ts
+++ b/src/app/content/shared/content.service.ts
@@ -57,7 +57,7 @@ export class ContentService {
     }
     if (environment.content_api.authenticationHeader) {
       const user = this.userService.getUser();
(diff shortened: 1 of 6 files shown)


src/app/content/shared/content.service.ts
```diff
@@ -47,11 +47,14 @@ export class ContentService {
...
     const urlParameters: string[] = [];
     if (webViewable) {
       urlParameters.push('rendition=Web');
     }
+    if (!isNullOrUndefined(disposition)) {
+      urlParameters.push('disposition=' + disposition);
+    }
     if (environment.content_api.authenticationHeader) {
       const user = this.userService.getUser();
       urlParameters.push('x-uw-act-as=' + user.actAs);
```
maximede: please get this url parameter from the config  `environment.content_api.authenticationHeader`

PR #42, opened 2017-11-01
implement typeahead

jfshenuw:


src/app/content/content-metadata/content-metadata.component.ts
```diff
@@ -37,9 +38,24 @@ export class ContentMetadataComponent implements OnInit, OnChanges, OnDestroy {
...
   private generateDisplayedMetadataGroup(): FormGroup {
     const group: any = {};
-    this.pageConfig.fieldsToDisplay.map(field => (group[field.name] = new FormControl('')));
+    this.pageConfig.fieldsToDisplay.map(field => {
+      if (field.type === 'typeahead' && field.options && field.options.length > 0) {
+        field.filteredOptions = new Observable<any[]>();
+        const fc = new FormControl('');
+        field.filteredOptions = fc.valueChanges
+          .startWith(null)
+          .map(x => (x ? this.filterOptions(x, field.options) : field.options.slice()));
```
agagne: is there a reason to return `field.options.slice()` rather than `field.options` in the event x evaluates to false?

jfshenuw: I got the code from Angular Material site. Either one would work. The one with slice makes a shallow copy, to be consistent with a user input, I guess.

PR #47, opened 2017-11-03
Feature/cab 2615

agagne:
- There are some style decisions that need to be made regarding the theme for checkbox, etc
- Currently you can drag multiple items to the file drop-zone, but it will only use the first to create. We have a separate ticket to allow uploading multiple files.
- I'm truncating the file names to 20 characters, this should probably be moved into a seperate pipe with variable length
- the file names that display should all be the same size, so they should be padded to 20 or whatever length we decide


src/app/content/content-view/content-view.component.html
```diff
@@ -22,6 +23,10 @@
       <img [src]="url | safeUrl"/>
     </div>
 
+
+    <div *ngSwitchCase="'unknown-dataURI'" layout="row">
+      <p >Unable to display Content Preview.</p>
```
maximede: This should probably use the customText pipe that James created

src/app/content/create-page/create-page.component.ts
```diff
@@ -0,0 +1,120 @@
...
+  prepareSaveContentItem(): ContentItem {
+    const formModel = this.createContentItemForm.value;
+    const updatedContentItem = new ContentItem();
+
+    // copy formModel updates into contentItem
+    for (const key of Object.keys(formModel.metadata)) {
+      updatedContentItem.metadata[key] = formModel.metadata[key];
+    }
+    updatedContentItem.metadata['ProfileId'] = this.config.profile;
+    updatedContentItem.metadata['Account'] = this.config.account + '/' + this.user.userName; // TODO: This should be done serverside
```
maximede: This will probably break non procard Accounts

PR #53, opened 2017-11-14
Implement CAB-2619 - New checkbox component

maximede:


src/app/shared/widgets/checkbox/checkbox-input.component.ts
```diff
@@ -0,0 +1,178 @@
...
+    });
+  }
+
+  ngOnInit(): void {
+    this.formGroup = this.fb.group({
+      internalCheckbox: new FormControl()
+    });
+
+    if (this.fieldConfig && this.fieldConfig.checkboxOptions) {
+      this._checkedValue = this.fieldConfig.checkboxOptions.checkedValue
```
agagne: should we be throwing an error if checkboxOptions are not provided, since they are required for the component to be usable?

PR #74, opened 2017-12-11
use AppName as header title

maximede:


src/app/routing/shared/config-resolver.service.ts
```diff
@@ -32,20 +32,28 @@ export class ConfigResolver implements Resolve<Config> {
...
         this.router.navigate(['/']);
         return null;
       }
     });
   }
 
-  getTenantNameSubject(): Observable<string> {
-    return this.tenantName$;
+  private getAppName(config) {
+    if (config.tenantFriendlyName) {
```
agagne: should this be config.appName?

maximede: yes indeed

PR #80, opened 2017-12-20
Feature/cab 2668

agagne:


src/app/content/edit-page/edit-page.component.html
```diff
@@ -13,7 +13,8 @@
           [contentItem]="contentItem"></app-content-pager>
       </span>
     </mat-toolbar>
-    <form [formGroup]="form">
+    <h2>{{pageConfig.pageName}}</h2>
+    <form [formGroup]="form" role="main">
```
maximede: `role="main"` is set on the main `div` in the `create-page-component`. Not sure what the best place is but these 2 pages should probably set up in the same way

src/app/search/search-box/search-box.component.html
```diff
@@ -1,15 +1,15 @@
-<div class="cs-search-box">
+<div class="cs-search-box" role="search">
     <div class="search-form-group">
-        <label *ngIf="pageConfig?.searchConfig?.label">{{pageConfig.searchConfig.label}}</label>
-        <mat-form-field>
-            <input matInput [(ngModel)]="searchModel.stringQuery" (search)="updateSearch()" (ngModelChange)="updateSearch()" placeholder="{{pageConfig?.searchConfig?.placeholder}}" type="search">
+        <label for="search-field" *ngIf="pageConfig?.searchConfig?.label">{{pageConfig.searchConfig.label}}</label>
+        <mat-form-field >
+            <input id="search-field" matInput [(ngModel)]="searchModel.stringQuery" (search)="updateSearch()" (ngModelChange)="updateSearch()" placeholder="{{pageConfig?.searchConfig?.placeholder}}" type="search">
```
maximede: should this be a `name` instead of `id` ?

agagne: No, for references an element id.

src/app/content/edit-page/edit-page.component.html
agagne: good catch, i've put them both in the form that contains all the the features of the page

PR #83, opened 2017-12-29
fix issue when ordering on number type field

maximede:


src/app/search/shared/search.service.spec.ts
```diff
@@ -213,4 +212,52 @@ describe('SearchService', () => {
       expect(payload.searchOrder.order).toEqual('desc');
     });
   });
+
+  it('shouldnt add .lowercase to number field Ordering to the query payload', () => {
+    const searchModel = new SearchModel();
+    searchModel.stringQuery = 'iSearch';
+    searchModel.order.term = 'myIntegerField';
+    searchModel.order.order = 'desc';
```
agagne: Should the default SearchOrder.order just be 'desc'?

PR #93, opened 2018-01-03
Feature/e2e tests 0.92

huiaic:
additional e2e tests for Create and Edit pages


e2e/create/create.po.ts
```diff
@@ -1,8 +1,9 @@
-import { browser, by, element } from 'protractor';
-import * as path from 'path';
+import {browser, by, element} from 'protractor';
 
 export class CreatePage {
   public pageUrl = '/demo/create';
+  public uploadFilePanel = element(by.id('drop-zone'));
+  public fileList = element.all(by.tagName('mat-list-item'));
```
agagne: We should probably avoid using selectors similar to 'mat-list-item' if possible, and probably add an id to the element your checking here. If we were to change the implementation of the file list it may break

PR #95, opened 2018-01-05
CAB-2713: Blank metadata takes value from prev/next item when navigating

agagne:


src/app/shared/widgets/options-input/options-input.component.ts
```diff
@@ -75,16 +75,12 @@ export class OptionsInputComponent implements ControlValueAccessor, MatFormField
...
-    }
+    this._value = value;
+    this.populateInitialValue(this.value);
   }
 
   private populateInitialValue(initialValue: string) {
-    if (!isNullOrUndefined(initialValue)) {
-      this.formGroup.controls['optionsForm'].patchValue(initialValue);
-    }
+    this.formGroup.controls['optionsForm'].patchValue(initialValue);
```
agagne: fields are initialized to '' so we should reset them rather than allow null/undefined

PR #183, opened 2018-08-23
Feature/cab 3137

agagne:
* updated to angular6 using the migration guide
* replace deprecated isNullOrUndefined & isUndefined from Node's util


.travis.yml
```diff
@@ -34,9 +34,9 @@ jobs:
       env:
         job=yarn e2e-local
     - stage: test
-      script: yarn build --prod --env=travis
+      script: yarn build --prod --configuration=travis
```
maximede: we'll need to make sure to update ansible too

agagne: looks like we don't specify the environment in ansible, just "yarn build -- --prod --base-href {{base_href}}" ..

PR #283, opened 2018-09-25
Feature/cab 3063 daterange add e2e tests

huiaic:


src/app/search/search-daterange-picker/search-daterange-picker.component.html
```diff
@@ -11,5 +11,6 @@
          (datesUpdated)="datesUpdated($event)"
          (rangeClicked)="datesUpdated($event)"
          autoApply="false"
+         id="date-range-input"
```
maximede: this might forbid us of creating multiple date range picker. Please use a class if your tests need to target it

PR #504, opened 2019-04-01
CAB-3517: Accessibility: UUI SearchPage Inadequate/Lack of Focus Indi…

agagne:
…cator

### DIFF
--- a/src/app/search/search-results/search-results.component.html
+++ b/src/app/search/search-results/search-results.component.html
@@ -43,7 +43,8 @@
     </ng-container>
 
     <mat-header-row *matHeaderRowDef="displayedColumns"></mat-header-row>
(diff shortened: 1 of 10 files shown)


src/app/search/search-results/search-results.component.html
```diff
@@ -43,7 +43,8 @@
     </ng-container>
 
     <mat-header-row *matHeaderRowDef="displayedColumns"></mat-header-row>
-    <mat-row *matRowDef="let row; columns: displayedColumns;"></mat-row>
+    <mat-row class="uw-highlight-row" *matRowDef="let row; columns: displayedColumns;" routerLink="../edit/{{row.id}}"
```
maximede: I don't think you are using the class

PR #528, opened 2019-04-19
Feature/a11y axe core

huiaic:
I fixed a bunch of missing aria-labels etc. Also added a test that checks for accessibility violations, but the test is disabled for now so it won't stop the build.


src/app/content/create-page/create-page.component.html
```diff
@@ -13,7 +13,7 @@
                   [fieldName]="'addFile'"
                   label="Add Files"
                   (fileSelected)="addFile($event)"
-                  [(formGroup)]="form"
```
agagne: Was this intended?

src/app/shared/widgets/header/header.component.html
```diff
@@ -8,7 +8,7 @@ <h1>
       <span *ngIf="!title"><a routerLink="/" class="cs-title">&nbsp;</a></span>
     </h1>
     <span class="cs-spacer"></span>
-    <button mat-icon-button [matMenuTriggerFor]="accountMenu" id="tenant-menu">
+    <button mat-icon-button [matMenuTriggerFor]="accountMenu" id="tenant-menu" aria-label="menu">
```
agagne: Should this label should probably be more descriptive than just menu?

Perhaps something like 'Account Menu'?

huiaic: yes, that is a better name! I'll change it

src/app/content/create-page/create-page.component.html
huiaic: yes, I got errors that it couldn't be binded otherwise. Also, this matches what's in the Edit page, and it didn't error out there.

PR #534, opened 2019-04-24
Allow to use a globally defined 'availableFields' in the config

maximede:


e2e/mocks/profile-api/demo.json
```diff
@@ -32,43 +32,111 @@
...
+      "displayType": "select",
+      "dynamicSelectConfig": {
+        "type": "test-type",
+        "labelPath": "metadata.Code"
+      }
+    },
+    {
+      "key": "Filer",
+      "label": "Filer",
+      "required": true
```
huiaic: do you really need this required? If yes, I'll need to update my e2e. Right now it's expecting all fields to be not required in demo. demo2 can have required fields.

e2e/mocks/profile-api/demo.json
```diff
@@ -32,43 +32,111 @@
...
+      "label": "Document Type"
+    },
+    {
+      "key": "ProfileId",
+      "label": "Profile Id"
+    },
+    {
+      "key": "PublishStatus",
+      "label": "Publish Status",
+      "required": true,
```
huiaic: required?

e2e/mocks/profile-api/demo.json
maximede: I don't, but this is what it was set to

PR #556, opened 2019-05-15
reset invalid value before save

jfshenuw:


src/app/shared/widgets/person-autocomplete/person-autocomplete.component.html
```diff
@@ -1,5 +1,5 @@
 <div [formGroup]="formGroup">
-  <input type="text"  matInput [formControlName]="internalFieldName" [matAutocomplete]="auto" [attr.aria-label]="placeholder">
+  <input type="text"  matInput [formControlName]="internalFieldName" [matAutocomplete]="auto" [attr.aria-label]="placeholder" (blur)="validate()">
```
maximede: Is `blur` triggered when you hit `enter` to save the form without moving out of the box ?

jfshenuw: blur was not triggered with my local test using the "demo" profile. However, the PanelClosing event was triggered, and the invalid value was reset.

src/app/shared/widgets/person-autocomplete/person-autocomplete.component.ts
```diff
@@ -111,15 +111,24 @@ export class PersonAutocompleteComponent extends _PersonAutocompleteComponentBas
...
+              this.filteredOptions = [person];
+            } else {
+              this.personService
+                .autocomplete(term)
+                .pipe(takeUntil(this.componentDestroyed))
+                .subscribe((results: PersonSearchResults) => {
+                  this.filteredOptions = results.content;
+                });
+            }
+          } else if (!term || term.trim().length === 0) {
```
agagne: This should probably just be an else, no need for the if conditional

PR #560, opened 2019-05-17
use display value for facet search filter

jfshenuw:


src/app/search/facets-box/facets-box.component.ts
```diff
@@ -78,10 +83,21 @@ export class FacetsBoxComponent implements OnInit, OnDestroy {
     });
   }
 
-  addFacetFilter(key: string, value: string, label: string) {
+  addFacetFilter(key: string, value: string, label: string, facetConfig?: FacetConfig) {
```
agagne: Can we add a unit test for this function?

PR #567, opened 2019-05-20
Feature/e2e

huiaic:


e2e/edit/edit.local-spec.ts
```diff
@@ -135,4 +135,28 @@ describe('Edit Page', () => {
...
   });
+
+  it('should display customized error message on save if custom message is configured', () => {
+    const editPageForBadItemWithCustomMsg = new EditPage('demo', 'unauthorized-updatedid');
+    editPageForBadItemWithCustomMsg.navigateTo();
+
+    editPageForBadItemWithCustomMsg.inputField.sendKeys('any text');
+    editPageForBadItemWithCustomMsg.saveButton.click();
+
+    const customizedErrMsgLabel = demoConfig.customText['error.content.update.403'].label;
```
maximede: would you want to add a test to expect `customizedErrMsgLabel` to not be blank ?

huiaic: good catch. fixed

PR #576, opened 2019-05-22
CAB-3461 - Announce autocomplete changes

maximede:
Also, refactor Student and Person autocomplete to :
- Only update the parent value when the internal field is updated with
an Object
- Rollback the field to the previous value when you tab out of the
autocomplete
- Only send a search query every 300ms ( debounce keystrokes)
- Show a loader when search is being executed


src/app/content/edit-page/edit-page.component.html
```diff
@@ -13,7 +13,7 @@
           [contentItem]="contentItem"></app-content-pager>
       </span>
     </mat-toolbar>
-    <h2>{{pageConfig.pageName}}</h2>
+    <h2>{{pageConfig.pageName}}}</h2>
```
jfshenuw: Is the extra "}" a typo or an intentional literal "}"?

maximede: typo, good catch

src/app/search/search-box/search-box.component.ts
```diff
@@ -84,18 +84,40 @@ export class SearchBoxComponent implements OnDestroy, OnInit {
...
-          .subscribe((results: SearchFilterableResult[]) => {
-            this.filteredOptions = results;
-          });
-      } else {
-        this.filteredOptions = [];
-      }
-    });
+    this.searchBoxEvent
+      .pipe(
+        takeUntil(this.componentDestroyed),
```
jfshenuw: I remember reading somewhere that "takeUntil" needs to be last one in a pipe. You may want to check it debounceTime needs to moved before takeUntil.

maximede: Good to know. I'll update that

src/app/shared/widgets/person-autocomplete/person-autocomplete.component.ts
```diff
@@ -103,37 +134,73 @@ export class PersonAutocompleteComponent extends _PersonAutocompleteComponentBas
   }
 
   private initInternalFormUpdateListener() {
+    // This will listen to every INTERNAL_FIELD_NAME value changes where the content is a non empty string.
```
agagne: `// This will listen to every INTERNAL_FIELD_NAME value changes where the content is a non empty string or a selected Person object.`?

src/app/shared/widgets/person-autocomplete/person-autocomplete.component.ts
```diff
@@ -103,37 +134,73 @@ export class PersonAutocompleteComponent extends _PersonAutocompleteComponentBas
...
     this.formGroup.controls[INTERNAL_FIELD_NAME].valueChanges
       .pipe(
-        startWith(null),
+        startWith(''),
+        filter(term => {
+          return typeof term === 'string' && term.length > 0;
+        }),
+        tap(() => (this.isLoading = true)),
+        debounceTime(300),
+        tap(term => console.log('searching for ' + term)),
```
agagne: might be clearer to use the term autocomplete rather than searching? ` console.log('autocomplete for ' + term))`

src/app/shared/widgets/person-autocomplete/person-autocomplete.component.ts
maximede: this is only true for the second subscription in this method ( there is a different comment for it )

src/app/shared/widgets/person-autocomplete/person-autocomplete.component.ts
```diff
@@ -103,37 +134,73 @@ export class PersonAutocompleteComponent extends _PersonAutocompleteComponentBas
...
+    const announcementMessage = 'selected ' + person.getNameAndEmployeeId();
+    console.log('liveAnnouncer : ' + announcementMessage);
+
+    this.liveAnnouncer.announce(announcementMessage, 'polite');
+  }
+
+  private announceSearchResults() {
+    let announcementMessage;
+    if (!this.filteredOptions || this.filteredOptions.length === 0) {
+      announcementMessage = 'No results';
```
agagne: Should we use standard language? the search component anounces "No results for autocomplete" not just "No results"

PR #614, opened 2019-06-17
CAB-2716: Display and announce warning when leaving page with unsaved data

agagne:


src/app/routing/shared/pending-changes.guard.ts
```diff
@@ -0,0 +1,18 @@
...
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class PendingChangesGuard implements CanDeactivate<ComponentCanDeactivate> {
+  canDeactivate(component: ComponentCanDeactivate): boolean | Observable<boolean> {
+    if (component.canDeactivate()) {
+      return true;
+    } else {
+      return confirm(
+        'You have unsaved changes. Press Cancel to go back and save these changes, or OK and your changes will be lost.'
```
maximede: Should this text come from `CustomTextUtilities.getCustomText` ( same as what `FacetsBoxComponent` is doing  )

agagne: good idea

PR #658, opened 2019-07-23
Cab 3633

jfshenuw:
Notes on the invalid ID issue. My test indicated that the "id" property was used by "aria-own" attribute, and the "id" attribute was used to initialize the "id" of the "owned" element.  A static "id" attribute, e.g. id="custom-input-1", worked. But dynamic values [id] set the "id" property only, and I need to use both [id] and [attr.id] to solve the invalid issue.

### DIFF
--- a/src/app/content/content-object-list/content-object-list.component.html
+++ b/src/app/content/content-object-list/content-object-list.component.html
@@ -3,9 +3,6 @@
     <mat-list-item *ngFor="let contentObject of contentObjects; let i = index" [ngClass]="selectedIndex > -1 && i == selectedIndex ? 'cs-selected' : ''">
  
         <p matLine (click)="selectObject(i)">
-            <button mat-icon-button>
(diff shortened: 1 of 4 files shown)


src/app/content/content-object-list/content-object-list.component.html
```diff
@@ -3,9 +3,6 @@
     <mat-list-item *ngFor="let contentObject of contentObjects; let i = index" [ngClass]="selectedIndex > -1 && i == selectedIndex ? 'cs-selected' : ''">
  
         <p matLine (click)="selectObject(i)">
-            <button mat-icon-button>
```
maximede: is the removal of the button expected ?

jfshenuw: yes, this is the first item from the ticket https://jira.cac.washington.edu/browse/CAB-3633:

Remove the cloud icon button next to the file. This button has no function. Or replace it with an image icon.

PR #672, opened 2019-08-01
Override webdriver-manager version to support chrome 76

maximede:


huiaic:
fix didn't work. Closing PR.

PR #770, opened 2019-12-10
get sections after receiving course from SWS

jfshenuw:


farmas:
I am not very familiar with this code base, so take my comments with a grain of salt:
- I am assuming that this changes are caused by some bugs found on the previous versions. One thing to consider is adding a test for the issues that were fixed with these changes. I would be useful for reviewers to understand what the problem was by looking at a spec.
- For remote calls, when the code subscribes to get the results, should there be some form of error handling? In general, is there a strategy of what the product should do if one of these requests fail? 

Thanks!
Federico

PR #810, opened 2020-01-22
Adds an aria-label to the mat-select component so that it displays the currently selected value when the select box is displayed.

farmas:
Fixes CAB-3803 (Selected value in drop downs not read by screen reader).

This change ONLY addresses the screen reader announcing the currently selected value of the dropdown. Interacting with the mat-select component with a screen reader is still a problem due to https://github.com/angular/components/issues/11083.

### DIFF
--- a/src/app/shared/widgets/options-input/options-input.component.html
+++ b/src/app/shared/widgets/options-input/options-input.component.html
@@ -1,5 +1,5 @@
 <div [formGroup]="formGroup">
-  <mat-select [formControlName]="internalFieldName" (selectionChange)="onSelect($event)">
+  <mat-select [formControlName]="internalFieldName" [aria-label]="ngControl?.value" (selectionChange)="onSelect($event)">
     <mat-option *ngFor="let option of this.options$| async" [value]="option.value">
       {{ option.displayValue }}
     </mat-option>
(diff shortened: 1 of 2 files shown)


src/app/shared/widgets/options-input/options-input.component.html
```diff
@@ -1,5 +1,5 @@
 <div [formGroup]="formGroup">
-  <mat-select [formControlName]="internalFieldName" (selectionChange)="onSelect($event)">
+  <mat-select [formControlName]="internalFieldName" [aria-label]="ngControl?.value" (selectionChange)="onSelect($event)">
```
farmas: Note that the display value is more complicated to get. Given the overall poor accessibility state of the mat-select component, I think that this is a good win for little complexity.

src/app/shared/widgets/options-input/options-input.component.spec.ts
```diff
@@ -57,3 +60,57 @@ describe('OptionsInputComponent', () => {
...
+    fixture = TestBed.createComponent(TestHostComponent);
+    componentElement = fixture.debugElement.queryAll(By.css('mat-select'))[0].nativeElement;
+
+    fixture.detectChanges();
+  });
+
+  it('should render aria-label attribute set to its option value', async(() => {
+    fixture.whenStable().then(() => {
+      fixture.detectChanges();
+      expect(componentElement.getAttribute('aria-label')).toEqual('optionOneValue');
```
farmas: Took me a while to get the unit test to work, even though the code setup is quite large, it helped me to better understand the internal workings of the component and it may be useful for other components where we need to test them via a host.

src/app/shared/widgets/options-input/options-input.component.html
huiaic: We might have to create another ticket to have the display value then.

PR #817, opened 2020-02-03
Create page should redirect to search page or clear form depending on 'Upload Another' checkbox status.

farmas:
Fixes CAB-2767


farmas:
Need to investigate why some e2e tests are failing.

PR #818, opened 2020-02-04
filter list of courses with course offerings

jfshenuw:


src/app/shared/widgets/course-input/course-input.component.spec.ts
```diff
@@ -68,7 +68,10 @@ class MockStudentService extends StudentService {
   getSections(year: string, quarter: string, curriculum: string, courseNumber?: string): Observable<any> {
```
agagne: @jfshenuw Would you add a unit test for when there is No Section associated with the course?

src/app/shared/widgets/course-input/course-input.component.ts
```diff
@@ -183,17 +183,49 @@ export class CourseInputComponent extends _CourseInputComponentBase
...
-          this.courseControl.patchValue(this.courseNumber);
+      const courses$ = this.studentService.getCourses(this.year, this.quarter, this.curriculumAbbreviation);
+      const sections$ = this.studentService.getSections(this.year, this.quarter, this.curriculumAbbreviation);
+
+      forkJoin(courses$, sections$).subscribe(([courses, sections]) => {
+        this.courseOptions = [];
+
+        if (courses && courses['Courses'] && sections && sections['Sections']) {
+          // get group sections by course
+          const secs = {};
```
farmas: If we have models for 'courses' and 'sections' consider casting this any objects into an Array of their respective types.

src/app/shared/widgets/course-input/course-input.component.ts
```diff
@@ -203,28 +235,33 @@ export class CourseInputComponent extends _CourseInputComponentBase
...
-          } else {
-            if (!this.sectionOptions.find(e => e['SectionID'] === this.section)) {
-              this.setSectionValue(this.sectionOptions[0]['SectionID']);
-            }
-            this.sectionControl.enable();
+      this.sectionOptions = [];
+      // get sections from associated course. no need to call DataAPI.
+      if (this.courseOptions && this.courseOptions.length > 0) {
+        this.courseOptions.forEach(c => {
+          if (this.courseNumber === c['CourseNumber']) {
```
farmas: Should it break out of this forEach when the first courseNumber is matched?

src/app/shared/widgets/course-input/course-input.component.ts
```diff
@@ -203,28 +235,33 @@ export class CourseInputComponent extends _CourseInputComponentBase
...
+        // disable section option
+        this.setSectionValue('');
+        this.sectionControl.disable();
+      } else if (this.sectionOptions.length === 1) {
+        if (this.section !== this.sectionOptions[0]) {
+          this.setSectionValue(this.sectionOptions[0]);
+        }
+        this.sectionControl.disable();
+      } else {
+        if (!this.sectionOptions.find(e => e === this.section)) {
```
farmas: Consider using Array.some() instead.

src/app/shared/widgets/course-input/course-input.component.spec.ts
```diff
@@ -117,10 +120,10 @@ describe('CourseInputComponent', () => {
   it('should update the model when a value is selected', () => {
```
farmas: From the code in course-input.component, there seem to be  a lot more conditions that could be checked. Consider expanding coverage for the scenarios when there are no section for course, single section per course and multiple sections per course.

src/app/shared/widgets/course-input/course-input.component.ts
```diff
@@ -183,17 +183,49 @@ export class CourseInputComponent extends _CourseInputComponentBase
...
-          this.courseOptions &&
-          this.courseOptions.length > 0 &&
-          !this.courseOptions.find(e => e['CourseNumber'] === this.courseNumber)
-        ) {
-          this.courseNumber = this.courseOptions[0]['CourseNumber'];
-          this.courseControl.patchValue(this.courseNumber);
+      const courses$ = this.studentService.getCourses(this.year, this.quarter, this.curriculumAbbreviation);
+      const sections$ = this.studentService.getSections(this.year, this.quarter, this.curriculumAbbreviation);
+
+      forkJoin(courses$, sections$).subscribe(([courses, sections]) => {
```
farmas: For my understanding, what happens when either of this requests fail?

src/app/shared/widgets/course-input/course-input.component.spec.ts
jfshenuw: It never happens, after courses without sections are filtered out.

src/app/shared/widgets/course-input/course-input.component.ts
jfshenuw: I think the whole thing would fail.

src/app/shared/widgets/course-input/course-input.component.ts
jfshenuw: No models are defined for them. They came from Student Web Services.

src/app/shared/widgets/course-input/course-input.component.ts
jfshenuw: changed "find" to "includes", which takes a simple parameter instead of a function.

src/app/shared/widgets/course-input/course-input.component.ts
jfshenuw: changed forEach to "some". break does not work within forEach function.

src/app/shared/widgets/course-input/course-input.component.ts
agagne: forkJoin will throw an error if the one of the inner observables throw an error.

src/app/shared/widgets/course-input/course-input.component.ts
agagne: Doesn't "some" change the logic?

src/app/shared/widgets/course-input/course-input.component.spec.ts
jfshenuw: I added a few "expect" to verify the filtered course list and section lists are what to be expected. There are 6 courses in the course list and only 3 of them have sections/offerings.

src/app/shared/widgets/course-input/course-input.component.ts
farmas: And what is the end user's experience when this happens? Is the reason why we don't handle errors is because they are globally taken care of somewhere else?

src/app/shared/widgets/course-input/course-input.component.ts
farmas: Consider adding typings for them, so that is better understood the kind of data that the client is dealing with.

src/app/shared/widgets/course-input/course-input.component.ts
```diff
@@ -183,17 +183,49 @@ export class CourseInputComponent extends _CourseInputComponentBase
...
+            this.courseNumber = '';
+          }
         }
-        this.updateSections(); // need to update sections when year/quarter changed, even if course remained the same
+
+        // need to update sections when year/quarter changed, even if course remained the same
+        this.updateSections();
       });
     } else {
       this.courseOptions = [];
```
farmas: Consider moving the 'else' statement to the top, right now is dangling at the bottom of the function and harder to understand that this comes from the very first else.

src/app/shared/widgets/course-input/course-input.component.ts
```diff
@@ -183,17 +183,49 @@ export class CourseInputComponent extends _CourseInputComponentBase
 
   private updateCourses() {
```
farmas: Consider adding comments that summarizes the intent of the work done by the function.

src/app/shared/widgets/course-input/course-input.component.ts
jfshenuw: not in this case, as there is only one entry for each courseNumber.

src/app/shared/widgets/course-input/course-input.component.ts
jfshenuw: There are pros and cons with strong typing. Additional types lead to boilerplate code and additional "touch points" for future changes.

src/app/shared/widgets/course-input/course-input.component.ts
jfshenuw: Generally speaking, we only handle runtime errors/exceptions if we can recover. If the backend server is down or has other problems, we could catch the exception and provide a specific error message, but we won't be able to get the course or sections.

src/app/shared/widgets/course-input/course-input.component.spec.ts
jfshenuw: The conditions you listed are already covered with existing test. There are 6 courses in the course list. One of them has 2 sections, two of them have one section, and three of them have no sections.

PR #819, opened 2020-02-04
Create page should redirect to search page or clear form depending on 'Upload Another' checkbox status.

farmas:
Fixes CAB-2767

### DIFF
--- a/e2e/create/create.local-spec.ts
+++ b/e2e/create/create.local-spec.ts
@@ -2,7 +2,6 @@ import { CreatePage } from './create.po';
 import { SearchPage } from '../search/search.po';
 import { browser, ElementArrayFinder } from 'protractor';
 import * as path from 'path';
-import { until } from 'selenium-webdriver';
 import { protractor } from 'protractor/built/ptor';
 import { ContentServicesUiPage } from '../app/app.po';
 
@@ -130,7 +129,7 @@ describe('Create Page for Demo', () => {
     page.clickAcceptAlert();
   });
 
-  it("should display Upload Another checkbox that is checked by default if setting 'uploadAnoter'=true", () => {
+  it("should display Upload Another checkbox that is checked by default if setting 'uploadAnother'=true", () => {
     expect(page.uploadAnotherCheckbox.isDisplayed());
     expect(page.uploadAnotherCheckbox.isSelected());
   });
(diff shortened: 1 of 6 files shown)


e2e/create/create.po.ts
```diff
@@ -10,11 +10,13 @@ export class CreatePage {
   public uploadAnotherCheckbox = element.all(by.name('uploadAnother')).get(1);
   public studentInput = element(by.css('app-student-autocomplete input'));
   public personInput = element(by.css('app-person-autocomplete input'));
+  public filerInput = element(by.name('Filer'));
   public saveButton = element(by.id('saveItem'));
+  public cancelButton = element(by.id('cancel'));
```
farmas: To be consistent between cancelButton and saveButton, I removed clickCancelButton in favor of this field.

e2e/create/create.po.ts
```diff
@@ -10,11 +10,13 @@ export class CreatePage {
...
   public clearButton = element(by.buttonText('clear'));
   public pdfViewer = element(by.tagName('pdf-viewer'));
   public formFields = element.all(by.tagName('mat-form-field'));
-  public requiredFields = this.formFields.all(by.css('[required=\'\']'));
+  public requiredFields = this.formFields.all(by.css("[required='']"));
```
farmas: Turns out this change is due to the pre-commit hook that runs 'prettier' and automatically reformats the quotes.

e2e/create/create.local-spec.ts
```diff
@@ -113,29 +126,26 @@ describe('Create Page for Demo', () => {
 
     expect(page.uploadFilePanel.isDisplayed());
 
-    page.clickCancelButton();
-    page.clickAcceptAlert();
-  });
-
-  it('should replace the correct file when 1 of many files is replaced', () => {
```
farmas: It is my belief that this test no longer applies, after a document has been saved, it is no longer possible to continue to interact with the file list (the user will either be redirected to search page or the file list will be cleared in preparation for next upload).

e2e/create/create.local-spec.ts
```diff
@@ -341,4 +353,19 @@ describe('Create Page for Demo2', () => {
       );
     });
   });
+
+  it("should display Upload Another checkbox that is un-checked by default if setting 'uploadAnoter'=false", () => {
```
huiaic: typo in title - "uploadAnother"

e2e/create/create.po.ts
huiaic: I thought we want single quotes only?

farmas: I thought so as well, but there is currently a husky pre-commit hook that runs this npm script:

`    
"pretty": "prettier --parser typescript --single-quote --print-width=120 --write",
`

And this is what is auto-changing this line, probably because there is a single quote inside the string.

e2e/create/create.local-spec.ts
farmas: T_T

PR #822, opened 2020-02-05
UUI Search page should display 4 digit year for Dates and DateTimes

farmas:
Fixes CAB-3686


src/app/shared/widgets/display-field/display-field.component.html
```diff
@@ -1,9 +1,9 @@
 <span *ngIf="field" [ngSwitch]="getFieldType()">
  <span *ngSwitchCase="'date'">
-  {{value | date: 'shortDate'}}
+  {{value | date: 'M/d/yyyy'}}
```
farmas: This is where I got the formats from: https://angular.io/api/common/DatePipe

PR #824, opened 2020-02-06
Adds clarifications to the tunneling instructions.

farmas:


README.md
```diff
@@ -36,6 +36,8 @@ yarn stop-mock-services
 
 ## Development server with backend services hosted in AWS EC2
 
+- **Important: Tunneling to the data-api does not work currently.**
```
agagne: why does tunneling to data-api not work?

README.md
```diff
@@ -36,6 +36,8 @@ yarn stop-mock-services
...
+- Make sure you have Node v13.
 - Install and configure the [AWS CLI](https://aws.amazon.com/cli/).
 - Install and configure the [AWS helper scripts](https://github.com/uw-it-edm/technical-operations/tree/master/aws-helper-scripts).
```
agagne: Why do we need Node v13?

README.md
farmas: I don't know. More configuration might be required, but I don't know enough about that API to know how to get it to work yet.

README.md
farmas: The script to substitutes variables in the template needs to load the 'fs' module from node (which is a ES module). Node 13 added that ability by default.

agagne: Should we have a concern about developing on a different version than we support in production? (odd versions are not LTS)

farmas: Created: https://jira.cac.washington.edu/browse/CAB-3887

PR #825, opened 2020-02-06
Adds e2e test to verify experience when user attempts to save a document but server returns error.

farmas:
Fixes CAB-3885

I am open to feedback on this approach. I tried several things, before deciding on this approach:
- Tried writing an angular HttpInterceptor that could return 500 error when dev conditions are met.
- Tried triggering the error from nginx using some of the other request headers.
- Getting the uploaded file name from nginx was too complicated and more time that I am willing to spend.


src/app/content/shared/content.service.ts
```diff
@@ -108,15 +108,19 @@ export class ContentService {
     return url;
   }
 
-  private buildRequestOptions() {
-    const requestOptionsArgs = {};
+  private buildRequestOptions(filename?: string) {
     const CONTENT_API = environment.content_api;
-    const headers = new HttpHeaders();
-    headers.set('Accept', 'application/json');
```
farmas: I am reading the source right, this Accept header is never sent, so I removed it (since it is the state of the system).

src/app/content/shared/content.service.ts
```diff
@@ -108,15 +108,19 @@ export class ContentService {
...
+    let headers = new HttpHeaders();
+
     if (CONTENT_API.authenticationHeader) {
       const user = this.userService.getUser();
-      requestOptionsArgs['headers'] = new HttpHeaders().append(CONTENT_API.authenticationHeader, user.actAs);
+      headers = headers.append(CONTENT_API.authenticationHeader, user.actAs);
     }
-    return requestOptionsArgs;
+
+    if (!!filename) {
```
agagne: is `!!filename` a typo?

src/app/content/shared/content.service.ts
```diff
@@ -108,15 +108,19 @@ export class ContentService {
...
+      headers = headers.append('x-uw-attachment', filename);
```
farmas: The gist is that I am sending the file name as a header so that nginx can read it and make decisions on it. One potential issue with this is that it makes the request bigger for no customer value, to mitigate this I can add an environment config that toggles this behavior.

farmas:
Using the file size as a way to make the test fail is probably a better solution for this scenario.

PR #852, opened 2020-03-02
Adds a 'currency' display type that can be used to show field values as currency on the search page.

farmas:
Fixes CAB-3901


src/app/shared/widgets/display-field/display-field.component.spec.ts
```diff
@@ -83,4 +89,48 @@ describe('DisplayFieldComponent', () => {
...
+
+      fixture.detectChanges();
+
+      const span = fixture.debugElement.query(By.css('span span'));
+      const el = span.nativeElement;
+
+      expect(el.innerHTML.trim()).toBe(expectedString);
+    };
+
+    it('should display text as currency', () => {
```
agagne: How would it display `12.34567` ?

farmas: $12.35

PR #853, opened 2020-03-03
Adds default left or right justification to cells of search page depending on the displayType.

farmas:
Fixes CAB-3902.

Summary: Row cell should always be aligned to its header cell. If it is left-aligned, left-most character of row cell should align to the left-most character of header cell. If it is right-aligned, right-most character of row cell should align to the right-most character of header cell.

Left alignment is automatically done by mat-table. Right alignment is done by matching the right-padding of the row cell and the header cell (taking into account the possibility of the up/down sorting arrow). The paddings of content inside the flex box are:

|   | Flex   | Left Padding  | Right Padding  |
|------|------|------|------|
| Header/Left/Sortable | Default | 0 | 18px (automatically mat-table up/down arrow) |
| Header/Left/Non-Sortable | Default |    0  |   inherit (automatically by mat-table)    |
| Header/Right/Sortable | Right |  0    |   18px (automatically mat-table up/down arrow)   |
| Header/Right/Non-Sortable | Right |   0   |   18px by CSS   |
| Content/Left/Sortable | Default   |  0    |  10px by CSS    |
| Content/Left/Non-Sortable | Default   |    0  |  10px by CSS    |
| Content/Right/Sortable | Right   |  0    |  18px by CSS     |
| Content/Right/Non-Sortable | Right  |  0    |   18px by CSS   |

### DIFF
--- a/src/app/search/search-results/search-results.component.css
+++ b/src/app/search/search-results/search-results.component.css
@@ -1,16 +1,20 @@
-.cell-type-date,
-.cell-type-dateTime,
-.cell-type-currency,
-.cell-type-number {
+.cell-align-right {
   /* Right align the content of header and row cells */
   justify-content: flex-end;
 }
 
-.mat-header-cell.cell-type-date > div,
-.mat-header-cell.cell-type-dateTime > div,
-.mat-header-cell.cell-type-currency > div,
-.mat-header-cell.cell-type-number > div {
-  /* Padding of right aligned header content so that it doesn't collide with the next header.
+.mat-header-cell .header-cell-content {
+  /* Align text to the left by default, this is needed because the sortable header is rendered
+     as a button, which by default centers its text. */
+  text-align: left;
+}
+
+.mat-header-cell.cell-align-right .header-cell-content {
+  text-align: right;
+}
+
+.mat-header-cell.cell-align-right > .header-cell-content {
+  /* This rule applies for the non-sortable header.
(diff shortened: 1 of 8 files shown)


src/app/search/search-results/search-results.component.html
```diff
@@ -32,14 +32,18 @@
...
         <span *ngSwitchDefault>
-         <mat-header-cell *matHeaderCellDef mat-sort-header>{{ col.label }}</mat-header-cell>
+         <mat-header-cell *matHeaderCellDef mat-sort-header [ngClass]="getCellCssClass(col)">
+           <div>{{ col.label }}</div>
+          </mat-header-cell>
         </span>
       </span>
 
-      <mat-cell *matCellDef="let element" class="cs-search-table-cell">
+      <mat-cell *matCellDef="let element" [ngClass]="getCellCssClass(col)">
```
farmas: Note: have to use ngClass instead of 'class' attribute interpolation, because ngClass respects the other classes dynamically added by mat-table.

src/app/search/search-results/search-results.component.css
```diff
@@ -1,3 +1,17 @@
-.cs-search-table-cell {
-  padding-right: 10px;
+.cell-type-date,
+.cell-type-dateTime,
+.cell-type-currency,
+.cell-type-number {
```
farmas: After further consideration, I am going to simplify these selectors into a single one so that it is easier to change alignment of other columns in the future.

PR #855, opened 2020-03-06
Switching between UUI profiles fails to display initial search results.

farmas:
Fixes CAB-3877.

### DIFF
--- a/e2e/mocks/content-services-api-mocks.conf
+++ b/e2e/mocks/content-services-api-mocks.conf
@@ -18,6 +18,9 @@ server {
     sub_filter '__current_host__' '$host';
     sub_filter_once off;
 
+    # Allows static pages to be served in response to POST requests.
+    error_page  405     =200 $uri;
+
     location ~ /profile/v1/app/.+/.+ {
         root /usr/share/nginx/html/profile-api/;
         rewrite ^/profile/v1/app/(.+)/(.+)$ /profile/v1/app/$2 last;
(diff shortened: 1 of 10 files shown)


e2e/mocks/content-services-api-mocks.conf
```diff
@@ -170,17 +173,10 @@ server {
...
-                return 204;
-            }
+    location ~ /search/v1/documents-(.+){
+            root /usr/share/nginx/html;
 
-            if ($request_method = 'POST') {
-                error_page 404 =200 /search-api/search.json;
-            }
-            try_files /$1.json /search-api/search.json;
+            try_files /search-api/$1.json =200 /search-api/search.json;
```
farmas: This change makes it possible to serve different files per profile in response to search requests.

e2e/mocks/profile-api/demo3.json
```diff
@@ -29,6 +29,14 @@
...
           "label": "Profile Id"
         },
+        {
+          "key": "DocumentType",
+          "label": "Document Type"
+        },
+        {
+          "key": "PublishStatus",
+          "label": "Publish Status"
+        },
```
farmas: The order of the fields is modified so that test can confirm that new set of columns was loaded after switching profiles.

e2e/search/search.po.ts
```diff
@@ -157,20 +158,42 @@ export class SearchPage {
...
+   */
+  waitForFirstRowValue(columnId: string, expectedText: string, timeoutMilliseconds: number = 5000): promise.Promise<any> {
+    return this.waitForFunc(
+      () => this.getResultsByColumn(columnId),
+      rows => rows && rows.length > 0 && rows[0].trim() === expectedText,
+      timeoutMilliseconds
+    ).then(() => expect(this.getDistinctResultsByColumn(columnId).then(rows => rows[0])).toEqual(expectedText));
+  }
+
+  private waitForFunc<T>(testFunc: () => promise.Promise<T>, predicate: (val: T) => boolean, timeoutMilliseconds: number = 5000): promise.Promise<any> {
```
farmas: The reason for this function is because Protractor/jasmine don't have a way to dynamically supply an error message when a 'wait' expression times out. The intent is for the test to be sequential (Protractor doesn't support async/await either T_T) but timeout errors to report meaningful error message with the last current value displayed.

huiaic: Protractor 6 supports async/await. We are on 5. Unfortunately it not backward compatible.

PR #862, opened 2020-03-12
Fixes so that dropdowns in course-input component announce selected values in screen reader.

farmas:
Fixes CAB-3888.

### DIFF
--- a/src/app/shared/widgets/course-input/course-input.component.html
+++ b/src/app/shared/widgets/course-input/course-input.component.html
@@ -1,20 +1,20 @@
 <mat-form-field>
-  <mat-select [formControl]="yearControl" [(value)]="year" [placeholder]="'Year'" (selectionChange)="onSelectYear($event)">
+  <mat-select [formControl]="yearControl" [(value)]="year" [placeholder]="'Year'" [aria-label]="'Year ' + year" (selectionChange)="onSelectYear($event)">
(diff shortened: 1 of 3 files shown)


src/app/shared/widgets/course-input/course-input.component.html
```diff
@@ -1,20 +1,20 @@
...
 </mat-form-field>
 <mat-form-field>
-  <mat-select [formControl]="quarterControl" [placeholder]="'Quarter'" (selectionChange)="onSelectQuarter($event)">
+  <mat-select [formControl]="quarterControl" [placeholder]="'Quarter'" [aria-label]="'Quarter ' + quarter" (selectionChange)="onSelectQuarter($event)">
     <mat-option *ngFor="let option of this.quarterOptions" [value]="option">{{option}}</mat-option>
   </mat-select>
 </mat-form-field>
 <mat-form-field>
-  <mat-select [formControl]="courseControl"  [required]="fieldConfig && fieldConfig.required" [placeholder]="'Course'" (selectionChange)="onSelectCourse($event)">
+  <mat-select [formControl]="courseControl"  [required]="fieldConfig && fieldConfig.required" [placeholder]="'Course'" [aria-label]="'Course ' + courseNumber + '-' + courseTitle" (selectionChange)="onSelectCourse($event)">
```
farmas: Aria label is set to "Course -" if no course is selected, I this is un-acceptable I can create a bound property that can handle that case.

PR #871, opened 2020-03-25
Adds ability to hide the filters so that the search table has more space to render.

farmas:
CAB-3824.

### DIFF
--- a/e2e/mocks/profile-api/demo4.json
+++ b/e2e/mocks/profile-api/demo4.json
@@ -1,0 +1,251 @@
+{
+  "account": "demo-account",
+  "tenant": "Demo4",
+  "appName": "Demo 4 - No Facets",
+  "customText": {
+    "addContentItemButton": {
+      "label": "Upload",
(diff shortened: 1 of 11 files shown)


huiaic:
I have some picky requests: 

1. Could you increase the color contrast for the button please. Make it a little darker - to at least #949494 per https://webaim.org/resources/contrastchecker/

1. The keyboard tabbing order is a little odd. It looks like it is on the same row as the table header, so I expected to be able to Shift-tab to it from ID column, but it's actually between the Facets and Page Navigator instead. Maybe move the button up to be on the same level as Page Nav?

farmas:
1. I removed the opacity style. The button now has the same fill color as the upload document icon.
2. In order for the button to be between the table and the page navigator I will have to move the button from search-page.component to search-results.component.

farmas:
- Added ability to keep the state of the filter panel visibility per session.
- Added ability to auto-collapse the facets area and remove the toggle button for a profile that has no facets.

PR #872, opened 2020-03-26
CAB-3920: UUI Search can not recover after bad request with special chars

agagne:
- recreate search subscription to searchModel$ on error
- Change searchModel$ to a Subject to prevent bad search parameters from creating an infinite loop when recreating search subscription on error
- don't pass searchModel$ observable to searchService

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,13 +26,10 @@
     "postinstall": "cd ./node_modules/protractor && npm i webdriver-manager@latest"
   },
   "lint-staged": {
-    "linters": {
       "*.ts": [
-        "npm run pretty",
+        "yarn run pretty",
         "git add"
       ]
-    },
-    "relative": true
   },
   "private": true,
   "dependencies": {
@@ -102,7 +99,7 @@
   "husky": {
     "hooks": {
       "pre-commit": "lint-staged",
-      "pre-push": "npm run validate"
+      "pre-push": "yarn run validate"
     }
   }
 }
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
     "lint": "ng lint content-services-ui --type-check",
     "e2e": "ng e2e",
-    "e2e-local": "$npm_execpath mock-services && ng e2e --protractor-config=protractor.conf.local.js --configuration=travis ; testStatus=$? ; yarn stop-mock-services ; exit $testStatus",
+    "e2e-local": "$npm_execpath mock-services && node --max_old_space_size=4096 'node_modules/@angular/cli/bin/ng' e2e --protractor-config=protractor.conf.local.js --configuration=travis ; testStatus=$? ; yarn stop-mock-services ; exit $testStatus",
     "validate": "npm-run-all --parallel test lint",
     "pretty": "prettier --parser typescript --single-quote --print-width=120 --write",
     "mock-services": "$npm_execpath start-mock-services",
--- a/package.json
+++ b/package.json
@@ -26,10 +26,13 @@
     "postinstall": "cd ./node_modules/protractor && npm i webdriver-manager@latest"
   },
   "lint-staged": {
+    "linters": {
       "*.ts": [
-        "yarn run pretty",
+        "npm run pretty",
         "git add"
       ]
+    },
+    "relative": true
   },
   "private": true,
   "dependencies": {
@@ -74,7 +77,7 @@
     "@types/node": "~13.7.2",
     "axe-webdriverjs": "^2.2.0",
     "codelyzer": "^5.2.1",
-    "husky": "^3.1.0",
+    "husky": "^4.2.3",
     "jasmine-core": "^3.5.0",
     "jasmine-reporters": "^2.3.2",
     "jasmine-spec-reporter": "^4.2.1",
--- a/package.json
+++ b/package.json
@@ -26,13 +26,10 @@
     "postinstall": "cd ./node_modules/protractor && npm i webdriver-manager@latest"
   },
   "lint-staged": {
-    "linters": {
       "*.ts": [
-        "npm run pretty",
+        "yarn run pretty",
         "git add"
       ]
-    },
-    "relative": true
   },
   "private": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "@types/node": "~13.7.2",
     "axe-webdriverjs": "^2.2.0",
     "codelyzer": "^5.2.1",
-    "husky": "^4.2.3",
+    "husky": "^3.1.0",
     "jasmine-core": "^3.5.0",
     "jasmine-reporters": "^2.3.2",
     "jasmine-spec-reporter": "^4.2.1",
--- a/src/app/search/display-search-page/display-search-page.component.ts
+++ b/src/app/search/display-search-page/display-search-page.component.ts
@@ -5,7 +5,7 @@ import { Title } from '@angular/platform-browser';
 import { SearchModel } from '../shared/model/search-model';
 import { SearchResults } from '../shared/model/search-result';
 import { SearchService } from '../shared/search.service';
-import { BehaviorSubject, Subject, Subscription } from 'rxjs';
+import { Subject, Subscription } from 'rxjs';
 import { DataService } from '../../shared/providers/data.service';
 import { Sort } from '../shared/model/sort';
 import { NotificationService } from '../../shared/providers/notification.service';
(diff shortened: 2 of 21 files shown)


src/app/user/tenant/tenant.component.spec.ts
```diff
@@ -1,13 +1,13 @@
...
-import { Config } from '../../core/shared/model/config';
-import { GlobalEventsManagerService } from '../../core/shared/global-events-manager.service';
+import {TenantComponent} from './tenant.component';
+import {NO_ERRORS_SCHEMA} from '@angular/core';
+import {ActivatedRoute} from '@angular/router';
+import {MaterialConfigModule} from '../../routing/material-config.module';
+import {ActivatedRouteStub} from '../../../testing/router-stubs';
+import {Title} from '@angular/platform-browser';
+import {Config} from '../../core/shared/model/config';
+import {GlobalEventsManagerService} from '../../core/shared/global-events-manager.service';
```
farmas: Did your editor automatically striped these spaces?

src/app/user/tenant/tenant.component.spec.ts
```diff
@@ -52,4 +52,5 @@ describe('TenantComponent', () => {
     const title = fixture.debugElement.injector.get(Title);
     expect(title.getTitle()).toBe('test-tenant');
   });
+
```
farmas: Unused extra line.

src/app/user/tenant/tenant.component.spec.ts
agagne: Yeah, it must've .. weird.. I'll look at the formatter sometime next week if I have time.

src/app/user/tenant/tenant.component.spec.ts
```diff
@@ -1,13 +1,13 @@
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import {async, ComponentFixture, TestBed} from '@angular/core/testing';
```
farmas: What actually changed in this file? Other than removing empty spaces was there another change?

agagne: No, I think I mistakenly modified the whitespace and shouldn't have pushed it.
I've reverted the file.

src/app/search/display-search-page/display-search-page.component.ts
```diff
@@ -79,25 +80,36 @@ export class DisplaySearchPageComponent implements OnInit, OnDestroy, AfterViewI
...
       this.searchModel$.next(this.initialSearchModel);
     });
   }
 
+  private addSearchSubscription() {
+    if (this.searchSubscription) {
+      this.searchSubscription.unsubscribe();
+    }
+    this.searchSubscription = this.searchModel$
+      .pipe(switchMap(searchModel => this.searchService.search(searchModel, this.pageConfig)))
```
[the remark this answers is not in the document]
agagne: This was removed on purpose, as the page does not have a searchbox updating the searchModel. 

However, I like the idea of moving the debounce & distinctUntilChange to searchService. This will standardize the way we handle search and be less error prone in the future if we were to add search functionality to another page.

src/app/search/display-search-page/display-search-page.component.ts
```diff
@@ -79,25 +80,36 @@ export class DisplaySearchPageComponent implements OnInit, OnDestroy, AfterViewI
...
+      .subscribe(
```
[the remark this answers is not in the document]
agagne: The test for this change may have been added after your comment. 

There is a test to see if search will break on search errors, it also confirms that the search is called for each update to the search model by counting the number of times searchService was called.

src/app/search/search-page/search-page.component.spec.ts
```diff
@@ -181,4 +186,65 @@ describe('SearchPageComponent', () => {
     expect(component.isLeftPanelVisible).toBeFalse();
     expect(component.isToggleLeftPanelButtonVisible).toBeFalse();
   });
+
+  describe('should recover from search error', () => {
+    const LongerThanSearchDebounceTime = 200;
+
+    let theSearchService: SearchService;
+    let searchSpy;
+    let searchModel: SearchModel;
```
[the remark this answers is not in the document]
agagne: good catch, i originally  had more tests in this section

src/app/search/display-search-page/display-search-page.component.ts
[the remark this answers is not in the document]
agagne: I take it back, if I were to add it in the searchService it would change the behavior and mean all searches would be in the same queue / have the same results which while true now doesn't seem like the right thing to do. I'll add the debounce & distinctUntilChange to the display page. 

That way behavior will remain the same, and it will be less errorprone to add additional searching to that page.

package.json
```diff
@@ -26,13 +26,10 @@
...
   "lint-staged": {
-    "linters": {
       "*.ts": [
-        "npm run pretty",
+        "yarn run pretty",
         "git add"
       ]
-    },
-    "relative": true
   },
```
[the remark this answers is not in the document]
agagne: @farmas Please review & approve https://github.com/uw-it-edm/content-services-ui/pull/873

yarn.lock
```diff
@@ -437,6 +437,11 @@
   resolved "https://registry.yarnpkg.com/@types/node/-/node-13.7.2.tgz#50375b95b5845a34efda2ffb3a087c7becbc46c6"
   integrity sha512-uvilvAQbdJvnSBFcKJ2td4016urcGvsiR+N4dHGU87ml8O2Vl6l+ErOi9w0kXSPiwJ1AYlIW+0pDXDWWMOiWbw==
 
+"@types/normalize-package-data@^2.4.0":
+  version "2.4.0"
+  resolved "https://registry.yarnpkg.com/@types/normalize-package-data/-/normalize-package-data-2.4.0.tgz#e486d0d97396d79beedd0a6e33f4534ff6b4973e"
+  integrity sha512-f5j5b/Gf71L+dbqxIpQ4Z2WlmI/mPJ0fOkGGmFgtb6sAu97EPczzbS3/tJKxmcYDj55OX6ssqwDAWOHIYDRDGA==
+
```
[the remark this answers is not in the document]
agagne: @farmas Please review @ approve https://github.com/uw-it-edm/content-services-ui/pull/873

src/app/search/display-search-page/display-search-page.component.ts
```diff
@@ -61,6 +64,11 @@ export class DisplaySearchPageComponent implements OnInit, OnDestroy, AfterViewI
     if (cachedSearch) {
       console.log('got cached search : ' + JSON.stringify(cachedSearch));
       this.searchModel$.next(Object.assign(new SearchModel(), cachedSearch));
+      this.changeDetectorRef.detectChanges();
+    } else {
+      // initialSearchModel should set after all components load, so that they will be subscribed to the searchModel$
+      this.searchModel$.next(this.initialSearchModel);
+      this.changeDetectorRef.detectChanges();
```
farmas: Hello Andrew, could you explain a little why this is needed? I am not clear what the issue is.

src/app/search/display-search-page/display-search-page.component.ts
```diff
@@ -1,31 +1,33 @@
-import { AfterViewInit, Component, OnDestroy, OnInit } from '@angular/core';
+import { AfterViewInit, ChangeDetectorRef, Component, OnDestroy, OnInit } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
 import { Config } from '../../core/shared/model/config';
 import { Title } from '@angular/platform-browser';
 import { SearchModel } from '../shared/model/search-model';
 import { SearchResults } from '../shared/model/search-result';
 import { SearchService } from '../shared/search.service';
-import { Subject, BehaviorSubject } from 'rxjs';
+import { BehaviorSubject, Subject, Subscription } from 'rxjs';
```
farmas: I believe BehaviorSubject is not used in this file.

src/app/search/search-page/search-page.component.spec.ts
```diff
@@ -181,4 +186,65 @@ describe('SearchPageComponent', () => {
     expect(component.isLeftPanelVisible).toBeFalse();
     expect(component.isToggleLeftPanelButtonVisible).toBeFalse();
   });
+
+  describe('should recover from search error', () => {
+    const LongerThanSearchDebounceTime = 200;
```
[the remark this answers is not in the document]
farmas: Looks like the debounce time was updated in display-search-page-component.spec but not in this one. Maybe an oversight?

src/app/search/display-search-page/display-search-page.component.ts
agagne: Great catch!

The page throwing an "Expression has changed after it was checked" error, due to the fact that the child page was was detecting the searchModel$ change that happened between the changeDetection.

I've pushed a change fixing the root cause, rather than forcing a change detection cycle.

PR #888, opened 2020-04-27
CAB-3801: Full screen mode does not limit keyboard navigation not in view

agagne:
CAB-3801: Full screen mode does not limit keyboard navigation not in view

- trap focus within content-view when fullscreen

### DIFF
--- a/src/app/shared/shared.module.ts
+++ b/src/app/shared/shared.module.ts
@@ -42,7 +42,7 @@ import { A11yModule } from '@angular/cdk/a11y';
     ReactiveFormsModule,
     FormsModule,
     PdfViewerModule,
-    A11yModule,
+    A11yModule
   ],
   exports: [
     CommonModule,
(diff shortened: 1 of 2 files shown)


src/app/shared/shared.module.ts
```diff
@@ -58,7 +67,7 @@ import { CourseInputComponent } from './widgets/course-input/course-input.compon
     ContentViewComponent,
     ContentToolbarComponent,
     CourseInputComponent,
-    ElasticsearchDateValidatorDirective
+    ElasticsearchDateValidatorDirective,
```
farmas, line 70: Un-needed comma?

farmas:
Could you add some explanation for reviewers of what the problem was and what that cdkTrapFocus does?

agagne:
The FocusTrap will constrain the tab to the content-view div when it is in fullscreen. This will resolve the bug where users who do not use a mouse will navigate to outside elements by 'tabing'.

**Expected keyboard navigation in dialog/modal**
https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-7

**Angular Material CDK Focus Trap**
https://material.angular.io/cdk/a11y/overview#example

> 
> **FocusTrap**
> The cdkTrapFocus directive traps Tab key focus within an element. This is intended to be used to create accessible experience for components like modal dialogs, where focus must be constrained.
> 
> This directive is declared in A11yModule.
> 
>  Example
> <div class="my-inner-dialog-content" cdkTrapFocus>
>   
> </div>
> This directive will not prevent focus from moving out of the trapped region due to mouse interaction.

PR #900, opened 2020-05-01
CAB-3801: add e2e test to check keyboard focus on full screen mode

huiaic:


e2e/edit/edit.local-spec.ts
```diff
@@ -5,8 +5,8 @@ import { ContentServicesUiPage } from '../app/app.po';
 import { until } from 'selenium-webdriver';
 import { protractor } from 'protractor/built/ptor';
 
-const getCurrentUrl = function() {
-  return browser.getCurrentUrl().then(url => {
+const getCurrentUrl = function () {
+  return browser.getCurrentUrl().then((url) => {
```
farmas, line 9: Consider not adding these parenthesis, single argument lambdas don't need them.

huiaic: linting added them automatically, I think.

PR #901, opened 2020-05-08
CAB-3889: Display error message when UUI Course Input Component cannot get a response from data-api.

farmas:


src/app/shared/widgets/course-input/course-input.component.spec.ts
```diff
@@ -133,6 +142,24 @@ describe('CourseInputComponent', () => {
     expect(component.courseOptions[2].sections.length).toBe(1);
     expect(component.sectionOptions.length).toBe(2);
   });
+
+  it('should raise a notification error if it fails to get courses', () => {
```
agagne, line 146: What will happen if there is no response? Do we have a timeout?

farmas: I have added a 10s timeout.

PR #902, opened 2020-05-08
CAB-3928: UUI feature update :Retaining the original file name on dow…

agagne:
…nloads

### DIFF
--- a/src/app/content/content-view/content-view.component.ts
+++ b/src/app/content/content-view/content-view.component.ts
@@ -1,6 +1,6 @@
 import { Component, Input, OnChanges, OnDestroy, OnInit, SimpleChanges, ViewChild } from '@angular/core';
 import { Subject } from 'rxjs';
-import { ContentService } from '../shared/content.service';
+import { ContentService, FileUrlParameters } from '../shared/content.service';
 
 import { ProgressService } from '../../shared/providers/progress.service';
 import { ContentObject } from '../shared/model/content-object';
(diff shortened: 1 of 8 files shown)


src/app/content/content-view/content-view.component.ts
```diff
@@ -96,7 +96,7 @@ export class ContentViewComponent implements OnInit, OnChanges, OnDestroy {
 
   private updateDownloadUrl(): void {
     if (this.contentObject && this.contentObject.itemId && this.contentObject.url !== '') {
-      this.downloadUrl = this.buildUrl(this.contentObject.itemId, true, 'attachment');
+      this.downloadUrl = this.buildUrl(this.contentObject.itemId, true, true, 'attachment');
```
farmas, line 99: Consider packaging the arguments into a type so that reader can better understand what each true means.

src/app/content/shared/content.service.spec.ts
```diff
@@ -43,7 +43,7 @@ describe('ContentService', () => {
     httpSpy = spyOn(http, 'get').and.returnValue(of(readResponse));
 
     const expectedUrl = environment.content_api.url + environment.content_api.contextV3 + '/item/123';
-    service.read('123').subscribe(result => {
+    service.read('123').subscribe((result) => {
```
farmas, line 46: Probably done by prettier, but consider relaxing the rule for parens in single argument lambdas.

src/app/content/shared/content.service.spec.ts
```diff
@@ -53,18 +53,41 @@ describe('ContentService', () => {
...
+    expect(
+      service.getFileUrl({
+        itemId: '123',
+        webViewable: false,
+        useOriginalFilename: true,
+        disposition: 'attachment',
+      })
+    ).toContain('disposition=attachment');
   });
   it('should to get the file url', () => {
```
huiaic, line 85:
```suggestion
  it('should get the file url', () => {
```

PR #905, opened 2020-05-13
fix course title issue with physics upload. delete invalid unit tests

jfshenuw:


src/app/shared/widgets/course-input/course-input.component.spec.ts
```diff
@@ -246,26 +246,6 @@ describe('CourseInputComponent with host', () => {
     });
   }));
 
-  it('should render aria-label set to placeholder for Course dropdown when course number is empty', async(() => {
```
agagne, line 249: @jfshenuw Did you mean to remove these unit tests?

farmas:
Consider adding a unit test that verifies the behaviour for this change. Looks like no unit test was set in place to catch the current bug.

PR #937, opened 2020-05-19
CAB-4019: Adds config setting to disable the replace file button.

farmas:


agagne:
Neat! I didn't know you could mark the PR as 'draft'

src/app/content/edit-page/edit-page.component.html
```diff
@@ -29,6 +29,7 @@ <h2>{{pageConfig.pageName}}</h2>
             <app-content-object-list
                     [contentItem]="contentItem"
                     [formGroup]="form"
+                    [disableFileReplace]="pageConfig && pageConfig.disableFileReplace"
```
agagne, line 32: Would it be better to add a `disableFileReplace` variable to the editPageComponenent, rather than adding the logic in the html?

farmas: Probably a good idea, it would enable TS checks with the model instead of the loosely typed expression in the template. I will make that change.

PR #943, opened 2020-05-28
CAB-3603: Adds a 'list-select' display type that uses an autocomplete component to allow users to filter the list of options.

farmas:
As part of this change I tried to reuse as much code from the OptionsInputComponent which was abstracted into a new service.


src/app/shared/widgets/options-autocomplete/options-autocomplete.component.ts
```diff
@@ -0,0 +1,197 @@
...
+  ngOnDestroy(): void {
+    this.componentDestroyed.next();
+    this.componentDestroyed.complete();
+  }
+
+  getOptionDisplayValue(option?: FieldOption) {
+    return option && option.displayValue;
+  }
+
+  optionListClosed() {
```
farmas, line 108: Note: I still need to add a test for this behavior.

PR #944, opened 2020-05-28
implement delete function based on user permissions

jfshenuw:


src/app/content/edit-page/edit-page.component.html
```diff
@@ -47,6 +47,15 @@ <h2>{{pageConfig.pageName}}</h2>
             <div class="cs-content-view-wrapper-insert">
               <app-content-view [contentObject]="contentObject" [allowPageByPageMode]="pageConfig?.allowPageByPageMode"></app-content-view>
             </div>
+            <div *ngIf="hasDeletePermission" class="button-row remove-button" style="margin: 2rem 25px 2rem 2rem;text-align: right">
```
farmas, line 50: Consider moving this style to the css file.

src/app/content/edit-page/edit-page.component.html
```diff
@@ -47,6 +47,15 @@ <h2>{{pageConfig.pageName}}</h2>
...
+              <mat-spinner *ngIf="deletePending" style="margin:0 auto;" aria-label="deletion in progress"></mat-spinner>
+              <button *ngIf="!deletePending" mat-raised-button aria-label="delete document"
```
farmas, line 52: Other buttons in app use matTooltip directive and also use upper case for the first letter of tooltip. Not sure if that is a standard though.

src/app/content/edit-page/edit-page.component.html
```diff
@@ -47,6 +47,15 @@ <h2>{{pageConfig.pageName}}</h2>
...
+                (click)="removeItem()" [color]="primary"
```
farmas, line 53: I recommend we use a different color for this potentially destructive operation, maybe "danger"?

src/app/content/edit-page/edit-page.component.ts
```diff
@@ -142,6 +147,44 @@ export class EditPageComponent extends ComponentCanDeactivateDirective implement
     this.submitPending = inProgress;
   }
 
+  getPermissions() {
```
farmas, line 150: - Consider adding a return type.
- Consider making private if possible.

src/app/content/edit-page/edit-page.component.ts
```diff
@@ -142,6 +147,44 @@ export class EditPageComponent extends ComponentCanDeactivateDirective implement
...
+        permissions = this.user.accounts[account];
+      }
+    }
+
+    return permissions;
+  }
+
+  getDeletePermission() {
+    const permissions = this.getPermissions();
+    return permissions && permissions.indexOf('d') >= 0;
```
farmas, line 165: Is the permission a single letter?

src/app/content/edit-page/edit-page.component.ts
```diff
@@ -142,6 +147,44 @@ export class EditPageComponent extends ComponentCanDeactivateDirective implement
...
+      const account = metadata && metadata['Account'];
+      if (account) {
+        permissions = this.user.accounts[account];
+      }
+    }
+
+    return permissions;
+  }
+
+  getDeletePermission() {
```
farmas, line 163: - Consider adding a return type.
- Consider renaming function so that it implies that it returns a boolean.
- Consider making it private.

src/app/content/edit-page/edit-page.component.html
```diff
@@ -47,6 +47,15 @@ <h2>{{pageConfig.pageName}}</h2>
             <div class="cs-content-view-wrapper-insert">
               <app-content-view [contentObject]="contentObject" [allowPageByPageMode]="pageConfig?.allowPageByPageMode"></app-content-view>
             </div>
+            <div *ngIf="hasDeletePermission" class="button-row remove-button" style="margin: 2rem 25px 2rem 2rem;text-align: right">
+              <mat-spinner *ngIf="deletePending" style="margin:0 auto;" aria-label="deletion in progress"></mat-spinner>
```
farmas, line 51: Should also check for submitPending, I am assuming we do not want to allow deletes while a save operation is on flight.

Same for the piece where we check for the submitPending above, we should also check for deletePending.

src/app/user/shared/user.service.ts
```diff
@@ -8,7 +8,8 @@ import { NotificationService } from '../../shared/providers/notification.service
 
 @Injectable()
 export class UserService {
-  private baseUrl: String = environment.profile_api.url + environment.profile_api.context;
```
farmas, line 11: - Are there any differences in terms of caching between the profile-api and content-api user endpoint?
- Do you know if this is the only consumer of profile-api user endpoint?

src/app/user/shared/user.ts
```diff
@@ -2,6 +2,7 @@ export class User {
   userName: string;
   actAs: string;
   userGroups: string[];
+  accounts: Map<string, string> = new Map();
```
farmas, line 5: Consider adding doc comments that explain what is the expected values of the Map.

farmas:
I think we should add a e2e test for this feature.

src/app/content/edit-page/edit-page.component.ts
jfshenuw: not necessarily.  Possible permissions are "r", "rw", "rwd", and "admin".

src/app/content/edit-page/edit-page.component.html
jfshenuw: OK, I will replace "primary" with "danger", which could change again after review by PM.

src/app/content/edit-page/edit-page.component.html
jfshenuw: moved style to .css file

src/app/content/edit-page/edit-page.component.html
jfshenuw: added matTooltip

src/app/content/edit-page/edit-page.component.ts
jfshenuw: good idea. made the change as you suggested.

src/app/content/edit-page/edit-page.component.ts
jfshenuw: added return type (boolean) and made it private. Regarding name, I started with "hasDeletePermission". But I need a variable to be used in html and the variable is named hasDeletePermission, and the function was renamed "getDeletePermission". I also thought about "hasDeleteRight" or "canDelete".

src/app/content/edit-page/edit-page.component.html
jfshenuw: made the change.

src/app/user/shared/user.service.ts
jfshenuw: I don't know if profile API is used by other app. 
Caching config for content API:
   <cache name="repository_user_info"
           maxEntriesLocalHeap="200"
           memoryStoreEvictionPolicy="LRU"
           timeToLiveSeconds="3600">
       <sizeOfPolicy maxDepth="1503" />
    </cache>

jfshenuw: I don't see cache config for profile api and I am not clear the details of caching in profile API.

src/app/user/shared/user.ts
jfshenuw: added sample value as a comment.

src/app/user/shared/user.service.ts
farmas: Oh, we may see some benefit of the content-api caching then :)

farmas:
Thanks Jay! I think the only thing that is missing is the e2e test.

src/app/content/edit-page/edit-page.component.ts
agagne: Would you add a comment of the possible permissions. It would be good to have context that when looking for the index of 'd', having it match 'admin' is the expected behavior and not a bug.

huiaic:
Converted to Draft mode so we don't accidentally merge. We should wait until we figure out if we want to offer Delete to all users with perms or add feature to turn it on/off.

src/app/content/edit-page/edit-page.component.ts
```diff
@@ -143,6 +148,48 @@ export class EditPageComponent extends ComponentCanDeactivateDirective implement
...
+    if (this.id && confirm('Do you want to delete the document with ID ' + this.id + '?')) {
+      this.deletePending = true;
+      this.contentService
+        .delete(this.id)
+        .pipe(takeUntil(this.componentDestroyed))
+        .subscribe(
+          (e) => {
+            this.deletePending = false;
+            const message = 'Deleted item ' + this.id;
+            alert(message); // snackBar did not work well just before navigating to another page
```
farmas, line 181: What was the problem with the snack bar? If I remember correctly the SnackBar has an afterOpened() observable that can be used to wait for the notification to be up.

jfshenuw: The message from snackbar showed up, but I could hardly see the message content before it was redirected to the search page, which could create confusion for end users. I tried to add a timeout to make the message displayed longer, but a given timeout may be short for one user and long for another. In the end, I felt that a simple alert message would work better.

PR #945, opened 2020-05-28
CAB-3869: Rename matTooltip for the remove file button and update element focus after file has been uploaded.

farmas:


farmas:
I don't think it is worth it to add an e2e test for this change. Let me know if you disagree and I can see how to add it.

PR #947, opened 2020-05-29
CAB-3603: Adds e2e and additional unit tests for OptionAutocompleteComponent.

farmas:
The new component is added to the demo profiles. Note that some tests automatically iterate over all metadata fields and some changes are added to take into account this new field.


src/app/shared/widgets/options-autocomplete/options-autocomplete.component.ts
```diff
@@ -158,7 +158,10 @@ export class OptionsAutocompleteComponent implements ControlValueAccessor, OnIni
 
     if (parentFieldConfig) {
       allOptions = concat(allOptions, this.parentControl.valueChanges.pipe(
+        distinctUntilChanged(),
         switchMap((newParentValue) => {
+          this.latestValidOption = null;
+          this.formGroup.controls[INTERNAL_FIELD_NAME].reset();
```
farmas, line 164: I think this is NOT an issue for the option-input-component, because when the option list is updated the UI automatically reflects the fact that the selected value no longer exists in the new list.

e2e/edit/edit.po.ts
```diff
@@ -8,13 +8,14 @@ export class EditPage {
   public profileUrl = `${browser.baseUrl}${this.profile}/edit`;
   public pageUrl = `${this.profileUrl}/${this.id}`;
   public pdfFilePath = path.resolve(__dirname, '../mocks/files/sample-file.pdf');
-  public inputField = element(by.id('mat-input-0'));
+  public inputField = element(by.className('mat-input-element'));
```
huiaic, line 11: This would match multiple elements and cause a warning message. So maybe refactor with `element.all()`, and in the tests, use `inputField.get(0)`.

PR #954, opened 2020-06-01
Feature/cab 3769 e2e

huiaic:


protractor.conf.local.js
```diff
@@ -25,6 +25,6 @@ exports.config = {
   },
   onPrepare() {
     require('ts-node').register({project: 'e2e/tsconfig.e2e.json'});
-    jasmine.getEnv().addReporter(new SpecReporter({spec: {displayStacktrace: true}}));
+    jasmine.getEnv().addReporter(new SpecReporter({spec: {displayStacktrace: 'pretty'}}));
```
huiaic, line 28: this was to remove the warning messages that came up. Looks like Jasmine's newer version changed the accepted param values.

e2e/create/create.local-spec.ts
```diff
@@ -113,6 +113,12 @@ describe('Create Page for Demo', () => {
     page.chooseFile(pdfFilePath);
 
     expect(page.pdfViewer.isDisplayed());
+
+    // CAB-3769
+    expect(page.addFileButton.getId()).toEqual(
```
farmas, line 118: This is a similar thing that I tried before. In file-upload.component.ts, if you comment out the line that sets the focus 
focus() {
    if (this.uploadButton) {
      //this.uploadButton.nativeElement.focus();
    }
  }

Does the test still pass?

farmas: I mention this, because the test passed for me and I think that if the test passes with and without the fix, then it is not really guarding against regressions.

huiaic: You're right, it didn't work :( 
I'll close the PR then.

farmas:
Should we still commit the part about the jasmine reporter?

PR #966, opened 2020-06-02
CAB-4050: 'Save and Upload Another' does not clear files list when multiple files are uploaded.

farmas:


e2e/mocks/content-services-api-mocks.conf
```diff
@@ -66,6 +66,10 @@ server {
     location ~ /content/v3/item/?(.*) {
         set $itemId $1;
 
+        if ($itemId = '') {
+            set $itemId $request_length;
+        }
```
farmas, line 71: This change makes the create item responses from having a valid id that can be used by the client.

PR #967, opened 2020-06-02
Addresses a race condition in e2e tests.

farmas:


e2e/create/create.local-spec.ts
```diff
@@ -243,11 +243,12 @@ describe('Create Page for Demo', () => {
...
 
-  it('should clear Employee value when value is not selected from auto complete and Save button is clicked', () => {
+  it('should clear Employee value when value is not selected from auto complete and Save and Upload Another button is clicked', () => {
     page.addFile(pdfFilePath);
     page.personInput.sendKeys('my employee');
-    page.saveButton.click();
+    page.saveAndResetButton.click();
 
+    browser.wait(isEmptyFileList(page.fileList), 5000);
     expect(page.getPersonValue()).toEqual('');
```
farmas, line 252: I believe this is race condition because after successful save the page will try to navigate back to search page and it may or may not have time to verify the status of the person field.

farmas: This is an example of the intermittent error:

    ✗ should clear Employee value when value is not selected from auto complete and Save button is clicked
      - Failed: stale element reference: element is not attached to the page document
        (Session info: headless chrome=83.0.4103.61)
        (Driver info: chromedriver=83.0.4103.39 (ccbf011cb2d2b19b506d844400483861342c20cd-refs/branch-heads/4103@{#416}),platform=Linux 4.15.0-1028-gcp x86_64)

huiaic: Makes sense. When the test was written, Saving did not trigger the navigation back to Search page.

PR #970, opened 2020-06-03
CAB-4054: Adds screen reader announcement when the filtered list updates.

farmas:


src/app/shared/widgets/options-autocomplete/options-autocomplete.component.ts
```diff
@@ -92,8 +92,16 @@ export class OptionsAutocompleteComponent implements ControlValueAccessor, OnIni
           }
           return true;
         })
-      )
+      ),
+      takeUntil(this.componentDestroyed)
```
farmas, line 96: I don't "think" this is needed for this observable chain, but looks like the best practice is to add the takeUntil to avoid leaks in the future.

src/app/shared/widgets/options-autocomplete/options-autocomplete.component.ts
```diff
@@ -175,6 +183,18 @@ export class OptionsAutocompleteComponent implements ControlValueAccessor, OnIni
...
 
+  private announceOptionsList(options: FieldOption[]) {
+    let message: string;
+
+    if (options && options.length > 0) {
+      message = `Filtered list has ${options.length} options.`;
+    } else {
+      message = 'Filtered list is empty.';
+    }
+
```
huiaic, line 194: I think we should keep the language consistent with the current person auto-complete. That one just says "Found x persons" or "No Results". Also when there is only 1 result, it's supposed to say the value, i.e "Found 1 option: [the option value]"

farmas: Ah ok, so something like:
- "Found no results".
- "Found one result: [option.displayValue]"
- "Found x results"

?

PR #980, opened 2020-06-09
OptionsAutocompleteComponent: Fixes bug where screen reader was announcing filter results after user selected an option.

farmas:


src/app/shared/widgets/options-autocomplete/options-autocomplete.component.spec.ts
```diff
@@ -111,14 +113,25 @@ describe('OptionsAutocompleteComponent', () => {
...
   });
 
+  it('should not announce option list if input control is not typed in', fakeAsync(() => {
+    component.registerOnChange(val => {});
+    component.filterInputControl.setValue(fieldOptions[1]);
+    fixture.detectChanges();
+
+    tick(1000);
+
+    expect(liveAnnouncerSpy.announce).not.toHaveBeenCalledWith('Found 3 results.', 'polite');
```
farmas, line 131: This test can be improved by asserting that the live announcer is not called at all.

PR #984, opened 2020-06-16
Bulk Update Prototype

farmas:
(Not to be merged)


farmas:
Prototype no longer needed.

PR #985, opened 2020-06-17
CAB-4053 (part 1): Adds capability to SearchResultsComponent to render in "FreezeResults" mode.

farmas:


src/app/search/shared/model/search-datasource.model.ts
```diff
@@ -13,9 +13,7 @@ export class SearchDataSource extends DataSource<ResultRow> {
   constructor(
     private searchModel$: Observable<SearchModel>,
     private searchResults$: Observable<SearchResults>,
-    private sort: MatSort,
-    private paginators: Array<MatPaginator>
```
farmas, line 17: Removed parameter not used anywhere.

src/app/search/search-results/search-results.component.spec.ts
```diff
@@ -7,46 +7,21 @@ import { SearchModel } from '../shared/model/search-model';
...
-import { ActivatedRouteStub } from '../../../testing/router-stubs';
-
-class MockDataService {
-  storage = ['123', '456'];
-}
-
-class RouterStub {
-  navigate(url: string) {
-    return url;
-  }
```
farmas, line 26: Removed all these mocks in favor of RouterTestingModule.

src/app/search/search-results/search-results.component.ts
```diff
@@ -43,6 +43,8 @@ export class SearchResultsComponent implements OnInit, OnDestroy {
   searchModel$: Observable<SearchModel>;
   @Input()
   searchResults$: Subject<SearchResults>;
+  @Input()
+  readOnly = false;
```
farmas, line 47: I admit that 'readOnly' might not be the best name for this input, I am open to suggestions.

src/app/search/search-results/search-results.component.html
```diff
@@ -1,5 +1,6 @@
 <div class="search-results-component">
   <mat-paginator #topPaginator
+                 *ngIf="!readOnly"
```
agagne, line 3: Why is pagination disabled for `readOnly`?

agagne: Nevermind, I think I understand.. when it is not 'readOnly' that means the search-results are in the 'bulk-edit' mode.

src/app/search/search-results/search-results.component.ts
agagne: Would you provide a comment that explains that readOnly is false, that the search-results are selectable for bulk-edit.

I'm not sure what the right wording is either, readOnly seems wrong.. theoretically I could see this for other uses such as bulk download, viewing multiple documents at once, etc. Maybe something like 'selectable' but that also doesn't seem like the right choice.

src/app/search/search-results/search-results.component.html
farmas: I think that is more evidence that 'readOnly' is not the right term. The intent is to kind of 'freeze' the table, meaning, a user can no longer paginate, sort or use any of the link navigation. This is used in two places: 1) in the search page, when the user selects which rows to include in the bulk edit operation and 2) in the bulk edit page, to just display the rows that are going to be updated.

Maybe call the property 'freezeNavigation' ??

src/app/search/search-results/search-results.component.ts
farmas: 'selectable' will be a separate toggle, the reason is that in the search page we need the table to be readOnly and selectable, but in the bulk edit page we need the table to be readonly and non-selectable.

src/app/search/search-results/search-results.component.html
agagne: 'freezeNavigation' or 'freezeResults' makes sense

PR #987, opened 2020-06-17
CAB-4052 (part 2): Adds capability to SearchResultsComponent for user to select rows and transmit selections with an event.

farmas:


src/app/search/search-results/search-results.component.html
```diff
@@ -21,6 +21,25 @@
     <!--- Note that these columns can be defined in any order.
         The actual rendered columns are set as a property on the row definition" -->
 
+    <ng-container matColumnDef="checked">
```
farmas, line 24: Adapted from this guide: https://material.angular.io/components/table/overview#selection

src/app/search/search-results/search-results.component.ts
```diff
@@ -79,7 +106,17 @@ export class SearchResultsComponent implements OnInit, OnDestroy {
...
-    this.dataSource = new SearchDataSource(this.searchModel$, this.searchResults$, this.sort);
+    const searchResultRows$ = this.searchResults$.pipe(
+      map(results => results.results),
+      tap(rows => this._resultRows = rows),
+      takeUntil(this.componentDestroyed)
+    );
+
+    this.dataSource = new SearchDataSource(this.searchModel$, searchResultRows$, this.sort);
+
+    this.selection.changed.pipe(debounceTime(100), takeUntil(this.componentDestroyed)).subscribe(() => {
```
farmas, line 117: Note: this debouncer is here to optimize the 'select-all/deselect-all' case, where a lot of selections will be done in quick succession and we only want to emit the event when they are done.

src/app/search/search-results/search-results.component.html
agagne: Thanks for including the link, it was useful during the review.

src/app/search/search-results/search-results.component.html
```diff
@@ -21,6 +21,27 @@
...
+      <mat-header-cell *matHeaderCellDef>
+        <mat-checkbox color="primary"
+                      [aria-label]="getHeaderSelectionLabel()"
+                      [matTooltip]="getHeaderSelectionAction()"
+                      (change)="$event ? toggleSelectAll() : null"
+                      [checked]="selection.hasValue() && areAllRowsSelected()"
```
agagne, line 30: @farmas Would it make sense to include `selection.hasValue()` in `areAllRowsSelected()`?

PR #989, opened 2020-06-18
mark form pristine after successful deletion

jfshenuw:


farmas:
- Why are there so many commits in this PR? Aren't all the other commits already in develop branch?
- There are code styling updates to files that are not related to the change, was this intentional?

PR #992, opened 2020-06-23
CAB-4063: First iteration of bulk edit page that allows users to change fields and update multiple documents at same time.

farmas:
Note: the text, instructions and position of UI elements haven't been locked down yet.

### DIFF
--- a/src/app/content/bulk-edit-page/bulk-edit-page.component.spec.ts
+++ b/src/app/content/bulk-edit-page/bulk-edit-page.component.spec.ts
@@ -46,7 +46,7 @@ function getConfigs(): { config: Config, pageConfig: BulkEditPageConfig } {
   return { config, pageConfig };
 }
 
-fdescribe('BulkEditPageComponent', () => {
+describe('BulkEditPageComponent', () => {
   let component: BulkEditPageComponent;
   let fixture: ComponentFixture<BulkEditPageComponent>;
   let activatedRoute: ActivatedRouteStub;
--- a/src/app/content/bulk-edit-page/bulk-edit-page.component.spec.ts
+++ b/src/app/content/bulk-edit-page/bulk-edit-page.component.spec.ts
@@ -38,7 +38,7 @@ function getConfigs(): { config: Config, pageConfig: BulkEditPageConfig } {
     Object.assign(new Field(), { key: 'field2', label: 'label 2' }),
     Object.assign(new Field(), { key: 'field3', label: 'label 3' }),
   ];
-  pageConfig.resultsFieldsToDisplay = pageConfig.fieldsToDisplay.slice(0);
+  pageConfig.resultsTableFieldsToDisplay = pageConfig.fieldsToDisplay.slice(0);
 
   const config = new Config();
   config.tenant = 'test-tenant';
@@ -46,7 +46,7 @@ function getConfigs(): { config: Config, pageConfig: BulkEditPageConfig } {
   return { config, pageConfig };
 }
 
-describe('BulkEditPageComponent', () => {
+fdescribe('BulkEditPageComponent', () => {
   let component: BulkEditPageComponent;
   let fixture: ComponentFixture<BulkEditPageComponent>;
   let activatedRoute: ActivatedRouteStub;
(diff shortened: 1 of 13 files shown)


src/app/content/bulk-edit-page/bulk-edit-page.component.ts
```diff
@@ -0,0 +1,193 @@
...
+import { BulkEditPageConfig } from '../../core/shared/model/bulk-edit-page-config';
+import { SearchPageConfig } from '../../core/shared/model/search-page-config';
+import { ResultRow } from '../../search/shared/model/result-row';
+import { SearchResults } from '../../search/shared/model/search-result';
+import { ContentItem, IContentItem } from '../shared/model/content-item';
+import { ContentService } from '../shared/content.service';
+import { DataService } from '../../shared/providers/data.service';
+
+const ROWS_LOCAL_STORAGE_KEY = 'bulk-edit-rows';
+const UPDATE_OPERATION_TIMEOUT = (3 * 60 * 1000); // 3 minutes
```
farmas, line 18: Is 3 minutes too much? too little?

src/app/content/bulk-edit-page/bulk-edit-page.component.ts
```diff
@@ -0,0 +1,193 @@
...
+    private _router: Router,
+    private _activatedRoute: ActivatedRoute,
+    private _formBuilder: FormBuilder,
+    private _titleService: Title,
+    private _contentService: ContentService,
+    private _dataService: DataService,
+    private _snackBar: MatSnackBar
+  ) {
+    const navigation: Navigation = _router.getCurrentNavigation();
+    this._rows = navigation && navigation.extras && navigation.extras.state && navigation.extras.state.selectedRows;
```
farmas, line 59: I left this using the navigation API instead of the native browser state, more due to the samples that I saw on angular docs than anything else. This is the only place that is used, let me know if you feel strongly about it I can change it.

src/app/content/content.module.ts
```diff
@@ -14,10 +16,11 @@ import { ContentToolbarComponent } from './content-toolbar/content-toolbar.compo
 import { ContentObjectListComponent } from './content-object-list/content-object-list.component';
 
 @NgModule({
-  imports: [FormsModule, SharedModule, ReactiveFormsModule, UserModule, PdfViewerModule],
+  imports: [FormsModule, SharedModule, ReactiveFormsModule, UserModule, PdfViewerModule, SearchModule],
```
farmas, line 19: We could consider moving the SearchResultsComponent to the Shared or Core module instead of referencing it like this. I opened this ticket to track: https://jira.cac.washington.edu/browse/CAB-4074

src/app/core/shared/model/bulk-edit-page-config.ts
```diff
@@ -0,0 +1,20 @@
...
+  /**
+   * Fields to display on the results table of the bulk edit page.
+   */
+  resultsFieldsToDisplay: Array<Field> = [];
+
+  /**
+   * List of keys used to reference fields from the 'availableFields' collection to display
+   *   on the results table. If defined, will take precedence over 'resultsFieldsToDisplay'.
+   */
+  resultsFieldKeysToDisplay: Array<string> = [];
```
farmas, line 19: These 2 config settings are new and are open to suggestions. Look at the demo.json file to see a sample of how the config looks like.

src/app/content/bulk-edit-page/bulk-edit-page.component.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    activatedRoute.testData = { config: getConfigs().config };
+
+    fixture = TestBed.createComponent(BulkEditPageComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  describe('on initial load', () => {
+    it('should save to local storage the rows received through navigation', () => {
```
farmas, line 90: Note: the idea behind this is that the page itself will store the rows it is working with in local storage. This enables the user to refresh the page without loosing progress.

src/app/content/bulk-edit-page/bulk-edit-page.component.ts
agagne: How long does it take to update ~50 documents with the bulk update endpoint?

src/app/content/shared/model/content-item.ts
```diff
@@ -1,7 +1,13 @@
 import { isNullOrUndefined } from '../../../core/util/node-utilities';
 import * as _ from 'lodash';
 
-export class ContentItem {
+export interface IContentItem {
```
agagne, line 4: I tend to be against Hungarian Notation, but I can't think of a better interface or class name that would remove the need.

src/app/core/shared/model/bulk-edit-page-config.ts
agagne: Maybe `resultsTableFieldsToDisplay` or `searchResultsTableFieldsToDisplay`?

src/app/shared/providers/data.service.ts
```diff
@@ -17,4 +17,14 @@ export class DataService {
   public set(key: string, object: any) {
     this.storage[key] = object;
   }
+
+  public setToLocalStorage(key: string, object: any): void {
```
agagne, line 21: Would it make sense to rename the other get/set methods to be more descriptive now that we have local storage as well?

src/app/content/shared/model/content-item.ts
farmas: I could not think of a better name for this interface either. It is unfortunate that we used classes for these kind of structures instead of interfaces or types, which I think is the more natural and preferred fit.

src/app/content/bulk-edit-page/bulk-edit-page.component.ts
farmas: I tried bulk updating 50 documents in dev environment and it took 30 seconds.

farmas: I am going to decrease this to 90 mins.

src/app/core/shared/model/bulk-edit-page-config.ts
farmas: I liked 'resultsTableFieldsToDisplay', I will make that change.

PR #993, opened 2020-06-24
CAB-4075: Adds E2E test for Bulk Update feature.

farmas:
There is only one E2E test here that verifies the whole user flow, other functionality is covered by unit tests.

Part of this PR is the beginnings of re-usable page objects: a page object for the SearchResultsComponent is created and used by several pages.

### DIFF
--- a/e2e/browserUtils.ts
+++ b/e2e/browserUtils.ts
@@ -1,0 +1,85 @@
+import { promise, browser, by, element, ElementFinder, ExpectedConditions } from 'protractor';
+
+/**
+ * Generic utility methods to interact with the browser.
+ */
+export class BrowserUtils {
+  /**
+   * Dismisses a browser's alert pop up.
+   * @param isAlertUnexpected Whether an alert is expected.
+   */
+  clickAcceptAlert(isAlertUnexpected: boolean = false) {
+    browser
+      .switchTo()
+      .alert()
+      .then((alert) => {
+        if (isAlertUnexpected) {
+          console.log('WARN: Unexpected alert left open from previous test. ');
+        }
+        alert.accept();
+      });
+  }
+
+  /**
+   * Waits for a snack bar to show with the given text.
+   * @param expectedText expected text to be contained in snack bar message.
+   * @param timeoutMilliseconds Timeout in milliseconds to wait for.
+   */
+  waitForSnackBarText(expectedText: string, timeoutMilliseconds: number = 5000): promise.Promise<any> {
+    const supplierFunc = () => element(by.className('mat-simple-snackbar'));
+
+    const visiblePromise = browser.wait(
+      ExpectedConditions.visibilityOf(supplierFunc()),
+      timeoutMilliseconds,
+      `SnackBar was not visible after ${timeoutMilliseconds} ms.`);
+
+    return visiblePromise.then(() => {
+      return this.waitForFunc(
+        () => supplierFunc().getText(),
+        text => (text || '').indexOf(expectedText) >= 0,
+        timeoutMilliseconds
+      ).then(() => expect(supplierFunc().getText()).toContain(expectedText));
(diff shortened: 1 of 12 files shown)


protractor.conf.local.js
```diff
@@ -25,6 +25,6 @@ exports.config = {
   },
   onPrepare() {
     require('ts-node').register({project: 'e2e/tsconfig.e2e.json'});
-    jasmine.getEnv().addReporter(new SpecReporter({spec: {displayStacktrace: 'pretty'}}));
+    jasmine.getEnv().addReporter(new SpecReporter({spec: {displayStacktrace: 'raw'}}));
```
farmas, line 28: I found the 'pretty' stack trace to be overwhelming and harder to parse, changed this to 'raw' for a more compact view.

e2e/search/search-results.po.ts
```diff
@@ -0,0 +1,112 @@
...
+      return Array.from(new Set(results));
+    });
+  }
+
+  /**
+   * Sorts results by the column header with the text specified.
+   * @param headerText The text of the column to sort the results by.
+   */
+  sortByHeaderText(headerText: string): promise.Promise<any> {
+    const publishStatusHeader = this.root.element(by.buttonText(headerText));
```
huiaic, line 52: Maybe rename `publishStatusHeader` to `columnHeader` to be generic?

e2e/search/search-results.po.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Gets the rows of the results table.
+   */
+  get rows(): ElementArrayFinder {
+    return this.root.all(by.css('app-search-results .mat-row'));
+  }
+
+  /**
+   * Gets the vales of each row cell of a given column.
```
huiaic, line 27:
```suggestion
   * Gets the values of each row cell of a given column.
```

e2e/search/search-results.po.ts
```diff
@@ -0,0 +1,112 @@
...
+  sortByHeaderText(headerText: string): promise.Promise<any> {
+    const publishStatusHeader = this.root.element(by.buttonText(headerText));
+    return publishStatusHeader.click();
+  }
+
+  /**
+   * Selects rows specified by index.
+   * @param indeces Array of row indeces to select.
+   */
+  selectRows(...indeces: number[]): promise.Promise<any> {
```
huiaic, line 60: My spell checker sense kicked in - should be "indices"

e2e/browserUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+    });
+  }
+
+  /**
+   * Waits for an element to contain the given text.
+   * @param elementOrSelector Element or selector to locate the element to test.
+   * @param expectedText Expected text.
+   * @param timeoutMilliseconds Timeout in milliseconds to wait for.
+   */
+  waitForElementText(elementOrSelector: string | ElementFinder, expectedText: string, timeoutMilliseconds: number = 5000): promise.Promise<any>  {
```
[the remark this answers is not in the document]
farmas: The problem with the way protractor designed wait-for-conditions is that it does not provide current value in the error message (and as far as I know, no way to supply the current value in the message). As in, if the condition fails, the log will just say "condition not met expected 'bar'", but for maintainability I would like the log to say "condition not met, the last value was 'foo' expected 'bar'", which is what this type of functions try to address.

e2e/search/search-results.po.ts
farmas: Oh english, why you so hard!

e2e/browserUtils.ts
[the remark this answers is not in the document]
huiaic: I see, that's too bad re Protractor limitations. Thanks for the explanation.

PR #997, opened 2020-06-26
CAB-4073: Add validation for cascading select fields so that child select is required if parent has a value set.

farmas:
I decided to make it configurable options on the content-metadata-component.

The E2E test is taking longer to implement, I will make a separate PR for that.


src/app/content/content-metadata/content-metadata.component.ts
```diff
@@ -90,4 +92,64 @@ export class ContentMetadataComponent implements OnInit, OnChanges, OnDestroy, A
...
+  reset(): void {
+    const metadataFormGroup = this.formGroup.get('metadata') as FormGroup;
+
+    this.formGroup.reset();
+
+    if (metadataFormGroup) {
+      Object.keys(metadataFormGroup.controls).forEach(key => {
+       metadataFormGroup.controls[key].setErrors(null);
+      });
+    }
```
farmas, line 119: I am not 100% sure why this happens. The problem is that when cross-field validation kicks in, the control has its [required] directive enabled, which then adds the error to the control's error array (so now there is the formgroup error and the control error), and for some reason calling form.reset() only clears the formgroup error, so this iteration goes and also clears any remaining control validation errors.

src/app/content/content-metadata/content-metadata.component.ts
```diff
@@ -90,4 +92,64 @@ export class ContentMetadataComponent implements OnInit, OnChanges, OnDestroy, A
     // return this.formGroup.controls['metadata'].controls[formControlName].hasError('required') ? 'You must enter a value' : '';
     return 'You must enter a value';
   }
+
+  isFieldControlRequired(field: Field): boolean {
```
jfshenuw, line 96: I was wondering if name could be "requiredFieldHasError" or something similar.

farmas: I see what you mean, the method name doesn't quite match what is being checked. I'll rename it.

PR #1002, opened 2020-06-29
CAB-4073: E2E tests for cascading field validators for Bulk Update.

farmas:
This pull request takes another step closer to have re-usable PageObjects for components.

### DIFF
--- a/e2e/bulk-edit/bulk-edit.local-spec.ts
+++ b/e2e/bulk-edit/bulk-edit.local-spec.ts
@@ -3,19 +3,21 @@ import { SearchPage } from '../search/search.po';
 import { ContentServicesUiPage } from '../app/app.po';
 import { BrowserUtils } from '../browserUtils';
 
-describe('Bulk Update Flow', () => {
-  const editPage = new BulkEditPage();
-  const searchPage = new SearchPage();
-  const utils = new BrowserUtils();
+const editPage = new BulkEditPage();
+const searchPage = new SearchPage();
+const utils = new BrowserUtils();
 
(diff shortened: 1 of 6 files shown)


e2e/page-objects/content-metadata.po.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  /**
+   * Gets the form field with the given label.
+   * @param label Label of the form field to locate.
+   */
+  getFormField(label: string): ElementFinder {
+    // XPath: Find element with tag 'mat-form-field' that has a descendant element with tag 'span'
+    //  and specified inner text.
+    const field = this.rootFinder.element(by.xpath(`//mat-form-field[.//span[text()='${label}']]`));
```
farmas, line 27: The reason for using xpath is to prevent WebDriver from acquiring intermediate WebElements that may go stale.

e2e/bulk-edit/bulk-edit.local-spec.ts
```diff
@@ -3,19 +3,21 @@ import { SearchPage } from '../search/search.po';
 import { ContentServicesUiPage } from '../app/app.po';
 import { BrowserUtils } from '../browserUtils';
 
-describe('Bulk Update Flow', () => {
-  const editPage = new BulkEditPage();
-  const searchPage = new SearchPage();
-  const utils = new BrowserUtils();
+const editPage = new BulkEditPage();
```
huiaic, line 6: How about renaming the variable to bulkEditPage to avoid confusion with the regular Edit page

PR #1016, opened 2020-07-06
CAB-3812: Allow multiple Data-API backed values to be selected in UUI.

farmas:
Notes:
- Adapted from: https://material.angular.io/components/chips/examples
- There are a lot of code style changes, this is because the husky pre-commit was not running on my system before (!!). Once I fixed that, the coding style changes kicked in.
- E2E tests will be done on a separate PR.

### DIFF
--- a/e2e/mocks/profile-api/demo.json
+++ b/e2e/mocks/profile-api/demo.json
@@ -67,6 +67,16 @@
       }
     },
     {
+      "key": "DataApiOptionMultiSelect",
+      "label": "DataApiOption with multi select",
+      "required": false,
+      "displayType": "multi-select",
+      "dynamicSelectConfig": {
+        "type": "test-type",
+        "labelPath": "label"
+      }
+    },
+    {
       "key": "Filer",
       "label": "Filer"
     },
--- a/src/app/content/bulk-edit-page/bulk-edit-page.component.ts
+++ b/src/app/content/bulk-edit-page/bulk-edit-page.component.ts
@@ -16,7 +16,7 @@ import { DataService } from '../../shared/providers/data.service';
 import { ContentMetadataComponent } from '../content-metadata/content-metadata.component';
 
 const ROWS_LOCAL_STORAGE_KEY = 'bulk-edit-rows';
-const UPDATE_OPERATION_TIMEOUT = (90 * 1000);
+const UPDATE_OPERATION_TIMEOUT = 90 * 1000;
 
 @Component({
   selector: 'app-bulk-edit-page',
@@ -101,23 +101,26 @@ export class BulkEditPageComponent implements OnInit, OnDestroy {
 
     this.isUpdatePending = true;
 
-    this._contentService.bulkUpdate(updatedRows).pipe(
-      timeout(UPDATE_OPERATION_TIMEOUT),
-      finalize(() => this.isUpdatePending = false),
-    ).subscribe(
-      response => {
-        const successes = (response && response.successes) || [];
-        const failures = (response && response.failures) || [];
-        this.removeRowsById(successes.map(row => row.id));
-        this.showPostUpdateMessage(successes, failures);
-      },
-      err => this._snackBar.open(`Bulk update operation failed to complete, please try again. Error: ${err.message || err}`, 'Dismiss')
-    );
+    this._contentService
+      .bulkUpdate(updatedRows)
+      .pipe(
+        timeout(UPDATE_OPERATION_TIMEOUT),
+        finalize(() => (this.isUpdatePending = false))
+      )
+      .subscribe(
+        (response) => {
+          const successes = (response && response.successes) || [];
+          const failures = (response && response.failures) || [];
+          this.removeRowsById(successes.map((row) => row.id));
+          this.showPostUpdateMessage(successes, failures);
+        },
+        (err) => this._snackBar.open(`Bulk update operation failed to complete, please try again. Error: ${err.message || err}`, 'Dismiss')
+      );
   }
 
   public removeRowsById(ids: string[]) {
     ids = ids || [];
-    this._rows = this._rows.filter(row => !ids.includes(row.id));
+    this._rows = this._rows.filter((row) => !ids.includes(row.id));
 
     const staticResults = new SearchResults();
     staticResults.results = this._rows;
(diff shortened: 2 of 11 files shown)


src/app/shared/widgets/options-autocomplete/options-autocomplete.component.ts
```diff
@@ -165,27 +182,78 @@ export class OptionsAutocompleteComponent implements ControlValueAccessor, OnIni
...
+  }
+
+  private addOptionToMultiSelect(option: FieldOption): void {
+    this.selectedOptions.push(option);
+    this.selectedOptionsChanged.next();
+    this.onChange(this.selectedOptions.map((opt) => opt.value));
+
+    this.filterInputControl.reset();
+    if (this.filterInputElement) {
+      this.filterInputElement.nativeElement.value = '';
```
farmas, line 199: There is something in the way MatChipList interacts with the filter input control, that just resetting the value does not clear the text in the UI (which is way the nativeElement needs to be cleared). See the https://material.angular.io/components/chips/examples for a similar pattern.

e2e/mocks/profile-api/demo.json
```diff
@@ -66,6 +66,16 @@
         "labelPath": "label"
       }
     },
+    {
+      "key": "DataApiOptionMultiSelect",
+      "label": "DataApiOption with multi select",
+      "required": false,
+      "displayType": "multi-select",
```
jfshenuw, line 73: I was wondering if a new UI type "multi-select" is redundant and if it is possible to derive "multi-select" from the profile type of the field (set in fieldMapping.yml). The UI type needs to be in sync with that in the fieldMapping and, if it could be derived, one could avoid the possible out-of-sync issue between UI config and fieldMap.

src/app/content/bulk-edit-page/bulk-edit-page.component.ts
```diff
@@ -101,23 +101,26 @@ export class BulkEditPageComponent implements OnInit, OnDestroy {
...
-      err => this._snackBar.open(`Bulk update operation failed to complete, please try again. Error: ${err.message || err}`, 'Dismiss')
-    );
+    this._contentService
+      .bulkUpdate(updatedRows)
+      .pipe(
+        timeout(UPDATE_OPERATION_TIMEOUT),
+        finalize(() => (this.isUpdatePending = false))
+      )
+      .subscribe(
+        (response) => {
```
jfshenuw, line 111: Would it be possible, in the future, to have format changes in its own commit and PR, to reduce the noise to feature PR?

e2e/mocks/profile-api/demo.json
farmas: I don't think UUI queries content-api for any fieldMapping information currently. Doesn't the same thing happens for the "date" type? (ie. it needs to be set in UUI config and fieldMaping).

I believe this is an improvement that we could make, but probably outside the scope of this PR.

src/app/content/bulk-edit-page/bulk-edit-page.component.ts
farmas: I apologize about that, it was in the middle of this work that I had to turn on prettier because Codacy started to reject my commits.

src/app/content/bulk-edit-page/bulk-edit-page.component.ts
```diff
@@ -161,17 +165,17 @@ export class BulkEditPageComponent implements OnInit, OnDestroy {
   }
 
   // tslint:disable-next-line:member-ordering
-  private static extractPageConfigs(config: Config): { editConfig: BulkEditPageConfig, searchConfig: SearchPageConfig } {
+  private static extractPageConfigs(config: Config): { editConfig: BulkEditPageConfig; searchConfig: SearchPageConfig } {
```
jfshenuw, line 168: Is semicolon the new list separator?

farmas: "Prettier" (correctly) identified this as a type declaration, similar to:

interface foo {
  key1: string;
  key2: string;
}

The linter must be more forgiving for inline types.

src/app/content/content-metadata/content-metadata.component.html
```diff
@@ -3,7 +3,18 @@
...
+            [placeholder]="fieldToDisplay.label"
+            [fieldConfig]="fieldToDisplay"
+            [parentControl]="formGroup.get('metadata').get(fieldToDisplay.dynamicSelectConfig?.parentFieldConfig?.key || '')"
+            [errorMessage]="getErrorMessage(fieldToDisplay.key)"
+            [required]="fieldHasRequiredError(fieldToDisplay)"
+            [id]="'custom-input-'+ i"
+            [attr.id]="'custom-input-'+ i">
+          </app-options-autocomplete>
+          <app-options-autocomplete *ngSwitchCase="'multi-select'"
+              [multiSelect]="true"
```
jfshenuw, line 17: Is [multiSelect] is the only difference between filter-select and multi-select? If it is, this new section may not be required. Instead, one could set multiSelect to false for filter-select.

farmas: Yes, that is the only difference. However ngSwitchCase doesn't support more than one value without hacks, open to suggestions.

src/app/shared/widgets/options-autocomplete/options-autocomplete.component.ts
```diff
@@ -69,46 +69,48 @@ export class OptionsAutocompleteComponent implements ControlValueAccessor, OnIni
...
-      takeUntil(this.componentDestroyed));
+      takeUntil(this.componentDestroyed)
+    );
 
-    this.filteredOptions$ = combineLatest(allOptions$, filter$).pipe(
+    this.filteredOptions$ = combineLatest(this.allOptions$, filter$, forceFilter$).pipe(
       map(([options, filter]) =>
         options.filter((option) => {
           const fieldOption = filter as FieldOption;
+          let keepOption = this.selectedOptions.findIndex((opt) => opt.value === option.value) < 0;
```
jfshenuw, line 92: could you add a comment the reason for "<0"? Are you initializing keepOption to true for non-selected options only here?

farmas: Will add a comment here.

src/app/content/content-metadata/content-metadata.component.html
jfshenuw: I was wondering if it could work with a new attribute, say "maxSelection", to existing selection type, without introducing a new multi-select type? maxSelection would default to 1 for backward compatibility and any number greater than 1 indicates multi-select.

farmas:
> t

Yeah, that could work, as long as product owner is ok with the changed UI and experience for single-select (for example, user would need to first 'remove' the option before being able to select a new one).

src/app/content/content-metadata/content-metadata.component.html
farmas: I see, yeah, that could work but it would change the experience for single-select in subtle ways. For example, it would require the user to remove a selection before being able to select a new one.

So far, the expectation for this feature is to add a new way to select multiple items, not to change the old way of selecting a single item. But it is something that we can bring up for sure.

farmas: Oh wait, you are talking about purely of config right? Something like:

```
    {
      "key": "DataApiOptionMultiSelect",
      "label": "DataApiOption with multi select",
      "displayType": "filter-select",
      "maxSelection": 1, 
      "dynamicSelectConfig": {
        "type": "test-type",
        "labelPath": "label"
      }
    },

```

I am fine with this approach as well. Let's talk about it during the next team meeting to see which way we want to go.

Do you think this is something that can be address in a subsequent PR?

jfshenuw: It is a config change. It could also be a design/implementation change, and I hope it would lead to fewer conditional code.

farmas:
I see what you mean, let's get some form of multiSelection to work so that the core experience can be tested in DEV and then I can make improvements on a) how to configure the field, and b) how single selection should work now that we have a chip list.

PR #1018, opened 2020-07-08
CAB-3812: Adds configuration option to 'filter-select' display field type in favor of the 'multi-select' display type.

farmas:
This was a suggestion that came from code review.

### DIFF
--- a/e2e/mocks/profile-api/demo.json
+++ b/e2e/mocks/profile-api/demo.json
@@ -58,7 +58,7 @@
     },
     {
       "key": "DataApiWithFilter",
-      "label": "DataApiOption show label with filter",
+      "label": "DataApiOption with filter",
       "required": false,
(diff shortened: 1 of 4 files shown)


e2e/mocks/profile-api/demo.json
```diff
@@ -68,9 +68,12 @@
     },
     {
       "key": "DataApiOptionMultiSelect",
-      "label": "DataApiOption with multi select",
+      "label": "DataApiOption with filter multi select",
       "required": false,
-      "displayType": "multi-select",
+      "displayType": "filter-select",
+      "filterSelectConfig": {
+        "multipleSelection": true
```
farmas, line 75: The original suggestion was 'maxSelection' but I thought that would create the wrong expectation since restricting the number of selections is a feature yet.

PR #1021, opened 2020-07-09
CAB-3812: Adds page objects and E2E test for filter select with multiple selection.

farmas:
Sadly, this changes includes formatting updates from prettier. They get changed at commit time, so its hard to know what's gonna happen.

### DIFF
--- a/e2e/edit/edit.local-spec.ts
+++ b/e2e/edit/edit.local-spec.ts
@@ -248,6 +248,25 @@ describe('Edit Page for Demo', () => {
 
     expect(page.getSnackBarText()).toEqual('There was an error deleting content item:Forbidden\nDismiss');
   });
+
+  it('should be able to add and remove options from filter select', () => {
+    // verify filter-select component loads chips from server and save button is disabled
+    const filterSelect = page.fields.getFilterSelectFormField('DataApiOption with filter multi select');
+    filterSelect.waitForSelectionCount(2);
+    expect(filterSelect.getSelectedValues()).toEqual(['val 1 label', 'val 3 label']);
+    expect(page.saveButton.isEnabled()).toBeFalsy();
+
+    // remove an option, verify save button becomes enabled
+    filterSelect.removeOptionAt(0);
+    filterSelect.waitForSelectionCount(1);
+    expect(filterSelect.getSelectedValues()).toEqual(['val 3 label']);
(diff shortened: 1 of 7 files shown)


e2e/page-objects/select-field.po.ts
```diff
@@ -49,7 +52,7 @@ export class SelectFieldPageObject {
   /**
    * Gets all the dropdown options.
    */
-  getOptions(): ElementArrayFinder {
+  get options(): ElementArrayFinder {
```
farmas, line 55: I changed this to match the new PO. In general, the pattern that I am trying to use is fields/properties to access 'Finders' (this seems to be the preferred approach from protractor because it delays finding the element until its used) and methods for operations that require WebDriver to acquire element references.

e2e/edit/edit.local-spec.ts
```diff
@@ -248,6 +248,25 @@ describe('Edit Page for Demo', () => {
 
     expect(page.getSnackBarText()).toEqual('There was an error deleting content item:Forbidden\nDismiss');
   });
+
+  it('should be able to add and remove options from filter select', () => {
```
huiaic, line 252:
```suggestion
  it('should be able to add and remove options from filter select with multi select', () => {
```

e2e/edit/edit.local-spec.ts
```diff
@@ -248,6 +248,25 @@ describe('Edit Page for Demo', () => {
...
+    filterSelect.waitForSelectionCount(1);
+    expect(filterSelect.getSelectedValues()).toEqual(['val 3 label']);
+    expect(page.saveButton.isEnabled()).toBeTruthy();
+
+    // add an option, verify selected options changed
+    filterSelect.addOptionAt(1);
+    filterSelect.waitForSelectionCount(2);
+    expect(filterSelect.getSelectedValues()).toEqual(['val 3 label', 'val 2 label']);
+  });
 });
```
huiaic, line 271: Maybe add a similar test to Create page to test that new values can added when the field is blank.

farmas: I decided to add the test for the filter-select that starts empty to the bulkeditpage instead, since that page has special handling for the emptyFormValidator. I did however add the filter-select field to the create page in the demo.json profile.

PR #1022, opened 2020-07-09
CAB-4084 fix an UI bug

jfshenuw:


src/app/shared/providers/student.service.ts
```diff
@@ -26,11 +26,15 @@ export class StudentService {
       - firstName
       - studentNumber */
   public autocomplete(term: string): Observable<StudentSearchResults> {
-    if (term && term.trim().length > 0) {
+    if (term && term.trim().length > 1) {
```
farmas, line 29: Consider adding a unit test, feels like it was missing.

jfshenuw: Not sure if a unit test is required here. No results for a one-character input is a limitation of SWS. I found that PWS does not have this limitation. A unit test to expect no results for a one-character might look strange.

PR #1024, opened 2020-07-10
CAB-4093: Add configuration setting to 'filter-select' field that sets the maximum selected options allowed.

farmas:


src/app/shared/widgets/options-autocomplete/options-autocomplete.component.spec.ts
```diff
@@ -266,6 +267,64 @@ describe('OptionsAutocompleteComponent', () => {
...
+      component.removeOptionFromMultiSelect(fieldOptions[1]);
+      fixture.detectChanges();
+      expect(autoComplete.options.first.disabled).toBeFalsy();
+    });
+
+    it('should announce the selection counts when adding and removing options', () => {
+      component.maxSelectionCount = 1;
+      component.optionSelected(fieldOptions[1]);
+      fixture.detectChanges();
+      expect(liveAnnouncerSpy.announce).toHaveBeenCalledWith('Selected display2 xy. Selected 1 of 1 options allowed.', 'polite');
```
agagne, line 322: What will the announcement be when there is no maximum?

agagne: nevermind, i see it just doesn't include the second sentence

farmas: If there is no maximum set, that last piece won't be appended and the message will just say 'Selected display2 xy.'

PR #1033, opened 2020-07-16
CAB-4044: UI improvements to ListFieldDisplayComponent.

farmas:


huiaic:
Have you tried it in IE? The `white-space` and `overflow` css properties are not supported by IE
https://www.w3schools.com/cssref/pr_pos_overflow.asp
https://www.w3schools.com/cssref/pr_text_white-space.asp

farmas:
Uh oh, I will give it a try in IE11 to see how it looks like.

PR #1045, opened 2020-07-20
Adds ability to override properties when using field references.

farmas:
The idea behind this feature is to be able to define the fields used for a profile once and re-use them for all pages. For EHS Radiation profile it will allow config to override just the one property that is different for the edit page:
```
fieldReferencesToDisplay: [
  { key: 'ReleaseDate', override: { disabled: true } },
  { key: 'DocumentType', override: { disabled: true } },
  'Building',
  'Floor'
],
```

### DIFF
--- a/src/app/core/shared/config.service.spec.ts
+++ b/src/app/core/shared/config.service.spec.ts
@@ -5,8 +5,10 @@ import { User } from '../../user/shared/user';
 import { ProgressService } from '../../shared/providers/progress.service';
 import { Config } from './model/config';
 import { environment } from '../../../environments/environment';
-import { Observable, of } from 'rxjs';
+import { of } from 'rxjs';
 import { TenantConfigInfo } from './model/tenant-config-info';
+import { Field } from './model/field';
+import { PageConfig } from './model/page-config';
 
 let httpSpy;
 let http: HttpClient;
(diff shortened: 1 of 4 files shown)


src/app/core/shared/config.service.spec.ts
```diff
@@ -58,10 +60,12 @@ describe('ConfigService', () => {
       expect(tenants.length).toEqual(2);
       expect(tenants[0].tenantName).toBe('demo');
       expect(tenants[1].tenantName).toBe('mech-eng-procard');
+
+      done();
```
farmas, line 64: Turns out these tests weren't running the expectations because they are inside of a promise, by adding the done function it forces jasmine to wait until the function is called.

src/app/core/shared/model/field.ts
```diff
@@ -4,6 +4,11 @@ import { DynamicSelectConfig } from './field/dynamic-select-config';
 import { CourseConfig } from './field/course-config';
 import { MultiSelectConfig } from './multi-select-config';
 
+/**
+ * Used to refer to a field by 'key'.
+ */
+export type FieldReference = string | { key: string; override?: Object };
```
farmas, line 10: This type is here so that in the simplest case the array will work with just a list of string keys, but then when we identify a field that needs an override, just that entry can be updated to an object without having to update all other strings.

src/app/core/shared/config.service.spec.ts
```diff
@@ -70,10 +74,80 @@ describe('ConfigService', () => {
...
+
+        done();
+      });
+    });
+
+    it('should override field properties specified on references array', (done: DoneFn) => {
+      pageConfig.fieldReferencesToDisplay = [{ key: 'field1', override: { disabled: true } }];
+
+      service.getConfigForTenant('demo').then((demoConfig) => {
+        const demoPage: PageConfig = demoConfig.pages['page1'];
```
huiaic, line 130: is it possible to add a 'page2' to verify that the override from page1 did not apply there?

PR #1068, opened 2020-08-28
CAB-4095: Upgrade Content-Services-UI to Angular 10

agagne:
We'll need to review the accessibility for the SearchPage's result table. The `mat-sort-header` no longer has an aria-label. The [MatSortHeaderIntl.sortButtonLabel ](https://material.angular.io/components/sort/api) has been deprecated in material 10

The [Angular Material PR](https://github.com/angular/components/pull/13297/files#diff-e5a01c54cda705d88178ecca63b2cb31R4) that removed this functionality has the comment

> We set the `tabindex` on an element inside the table header, rather than the header itself,
> because of a bug in NVDA where having a `tabindex` on a `th` breaks keyboard navigation in the
> table (see https://github.com/nvaccess/nvda/issues/7718). This allows for the header to both
> be focusable, and have screen readers read out its `aria-sort` state. We prefer this approach
> over having a button with an `aria-label` inside the header, because the button's `aria-label`
> will be read out as the user is navigating the table's cell (see #13012).


src/app/content/content-metadata/content-metadata.component.html
```diff
@@ -102,14 +102,15 @@
...
           <mat-form-field *ngSwitchDefault>
-            <input matInput [formControlName]="fieldToDisplay.key"
-                   [name]="fieldToDisplay.key"
-                   [placeholder]="fieldToDisplay.label"
-                   [attr.aria-label]="fieldToDisplay.label"
-                   [required]="fieldHasRequiredError(fieldToDisplay)"
-            >
-            <mat-icon *ngIf="fieldToDisplay.disabled" matSuffix class="disabled-icon" matTooltip="Read-only">lock</mat-icon>
-            <mat-error> {{getErrorMessage(fieldToDisplay.key)}}</mat-error>
+              <mat-label>{{fieldToDisplay.label}}</mat-label>
```
agagne, line 105: Needed to add the `mat-label`, to retain the `placeholder` attribute, without the field only included a `data-placeholder`

src/app/search/search-results/search-results.component.spec.ts
```diff
@@ -92,14 +92,6 @@ describe('SearchResultsComponent', () => {
     expect(component.sort.direction).toBe('desc');
   });
 
-  it('should use the field label for the aria attribute of sorting button', () => {
```
agagne, line 95: Removed this test as the arial-label is no longer generated. (See initial PR comment)

src/app/search/search-results/search-results.component.spec.ts
```diff
@@ -112,13 +104,13 @@ describe('SearchResultsComponent', () => {
   });
 
   it('should disable all sorting headers when freezeResults=true', () => {
-    let sortHeaders = fixture.debugElement.queryAll(By.css('.mat-sort-header-button[disabled=true]'));
+    let sortHeaders = fixture.debugElement.queryAll(By.css('.mat-sort-header-disabled'));
```
agagne, line 107: the generated html no longer includes `disabled=true`, switched to looking for the `.mat-sort-header-disabled` to identify if the sorting has been disabled

src/app/content/bulk-edit-page/bulk-edit-page.component.spec.ts
```diff
@@ -95,8 +95,8 @@ describe('BulkEditPageComponent', () => {
 
   describe('on initial load', () => {
     it('should save to local storage the rows received through navigation', () => {
-      const router = TestBed.get(Router);
-      spyOn(router, 'getCurrentNavigation').and.returnValue({ extras: { state: { selectedRows: testRows } } });
+      const router = TestBed.inject(Router);
```
agagne, line 98: The new `TestBed.inject` has the same as `TestBed.get`, but it is also typesafe. This change required the addition of `as any` to the spy for that reason.

package.json
```diff
@@ -74,31 +74,31 @@
     "codelyzer": "^6.0.0",
     "coveralls": "^3.1.0",
     "husky": "^4.2.5",
-    "jasmine-core": "^3.6.0",
+    "jasmine-core": "~3.5.0",
```
farmas, line 77: I see a couple of downgrades in this file. Do you know if we need to lock the dependencies to these versions? I ask in case I need to update the Renovate config.

src/app/content/bulk-edit-page/bulk-edit-page.component.spec.ts
farmas: Could you explain why the need for "as any"? Is it because getCurrentNavigation returns a type that is incompatible with the returned value from the mock?

src/app/content/content-metadata/content-metadata.component.html
farmas: Could you provide more context of what happened here? Given that we are putting the label for the input here, do we still need to pass the [placeholder] in the binding of the input below?

src/app/search/search-results/search-results.component.ts
```diff
@@ -219,12 +213,6 @@ export class SearchResultsComponent implements OnInit, OnDestroy {
       fieldLabelMap[field.key] = field.label;
     }
 
-    this.matSortService.sortButtonLabel = (id) => {
-      const label = fieldLabelMap[id] || id;
-
-      return `Change sorting for ${label}`;
```
farmas, line 225: Where did this text ended up after the update?

src/app/search/search-results/search-results.component.ts
```diff
@@ -97,13 +97,7 @@ export class SearchResultsComponent implements OnInit, OnDestroy {
   @ViewChild(MatSort, { static: true })
   sort: MatSort = new MatSort();
 
-  constructor(
-    private route: ActivatedRoute,
-    private router: Router,
-    private data: DataService,
-    private liveAnnouncer: LiveAnnouncer,
-    private matSortService: MatSortHeaderIntl
```
farmas, line 105: If this is no longer used, consider removing the using statement.

src/app/shared/widgets/options-autocomplete/options-autocomplete-component-base.directive.ts
```diff
@@ -12,7 +12,8 @@ const INTERNAL_FILTER_CONTROL_NAME = 'internalFilterControl';
 /**
  * Base class for components that have an autocomplete filter.
  */
-export class OptionsAutocompleteComponentBase implements OnInit, OnDestroy, ControlValueAccessor {
+@Directive()
```
farmas, line 15: For my own understanding, is the Directive decorator gone in Angular 10?

src/app/shared/widgets/options-autocomplete/options-autocomplete-component-base.directive.ts
```diff
@@ -12,7 +12,8 @@ const INTERNAL_FILTER_CONTROL_NAME = 'internalFilterControl';
...
+export class OptionsAutocompleteComponentBaseDirective implements OnInit, OnDestroy, ControlValueAccessor {
```
farmas, line 16: Why is this class now a Directive? It was meant to be the base class for 2 components. Is this something different in Angular 10?

src/app/content/content-metadata/content-metadata.component.html
agagne: The placeholder is creating the 'data-placedholder' attribute, but not the 'placeholder'. I added the label to retain the 'placeholder' within the input html field.

This change was required to prevent the removal of the 'placeholder' attribute.

src/app/content/bulk-edit-page/bulk-edit-page.component.spec.ts
agagne: The object defined in `returnValue` does not include all of the required fields, as the mock doesn't require those fields. I've cast to `any`.

src/app/search/search-results/search-results.component.ts
agagne: This text is no longer generated by angular, please see the initial comment for this PR. I reference the Angular team's reasoning. @huiaic @farmas If we decide that this is insufficient to our accessibility requirements, we can add custom custom code to generate this.

src/app/search/search-results/search-results.component.ts
agagne: @farmas What statement are you referring to?

src/app/shared/widgets/options-autocomplete/options-autocomplete-component-base.directive.ts
agagne: This is required for Angular 10's Ivy Compiler. 

https://angular.io/guide/migration-undecorated-classes

All base classes for components need to be annotated as either a @Directive() or @Componenent().

src/app/content/content-metadata/content-metadata.component.html
farmas: The confusing part is why a mat-label component is now needed for this input element, but not for the other components that also have input elements.

It is not necessary to get to the bottom of it.

src/app/content/bulk-edit-page/bulk-edit-page.component.spec.ts
farmas: aaah, I get it now. Thanks.

src/app/search/search-results/search-results.component.ts
farmas: MatSortHeaderIntl, is it used anywhere in the file anymore?

src/app/shared/widgets/options-autocomplete/options-autocomplete-component-base.directive.ts
farmas: Ooh, this is good to know. Thanks for the link.

farmas (dismissed):

e2e/page-objects/search-results.po.ts
```diff
@@ -42,11 +42,11 @@ export class SearchResultsPageObject {
   }
 
   /**
-   * Sorts results by the column header with the text specified.
-   * @param headerText The text of the column to sort the results by.
+   * Sorts results by the column header with the 'mat-column-<columnKey>' specified.
+   * @param columnKey The text of the column to sort the results by.
```
farmas, line 46: Update the comments for the columnKey argument.

PR #1076, opened 2020-09-01
CAB-4023: UUI: Move documentation from github wiki to source code.

agagne:


docs/CONFIGURATION.md
```diff
@@ -0,0 +1,215 @@
+This page will describe each configuration section available for the application.
+
+# Content Configuration
+
+TBD
```
farmas, line 5: Consider removing TBD from this file. They tend to just stay there forever.

PR #1127, opened 2020-11-24
CAB-4024: Adds unit and integration tests for auto-focus on search box.

farmas:
Test strategy:

Instead of verifying each individual navigation case, the primary goal is to verify that the search page puts focus on the search-box if the right navigation state is supplied by the source of navigation.

Additionally, verify the special case when exiting bulk edit mode the search page puts focus on the search-box.

Navigation sources can be tested via unit tests, and I updated the bulk-edit page unit test to check that the navigation state is passed.

### DIFF
--- a/e2e/browserUtils.ts
+++ b/e2e/browserUtils.ts
@@ -39,7 +39,6 @@ export class BrowserUtils {
    * Waits for a snack bar to show with the given text.
    * @param expectedText expected text to be contained in snack bar message.
    * @param timeoutMilliseconds Timeout in milliseconds to wait for.
-   * @param timeoutMilliseconds Timeout in milliseconds to wait for.
    */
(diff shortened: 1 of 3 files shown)


e2e/browserUtils.ts
```diff
@@ -20,23 +20,40 @@ export class BrowserUtils {
...
+      (actualId) => actualId === expectedElementId,
+      timeoutMilliseconds
+    ).then(() => expect(getActiveElementIdFunc()).toBe(expectedElementId));
+  }
+
   /**
    * Waits for a snack bar to show with the given text.
    * @param expectedText expected text to be contained in snack bar message.
    * @param timeoutMilliseconds Timeout in milliseconds to wait for.
+   * @param timeoutMilliseconds Timeout in milliseconds to wait for.
```
huiaic, line 42: looks like duplicated comments?

huiaic (changes requested):

huiaic (approved):

PR #1145, opened 2020-12-15
CAB-4199: Disable bulk edit and upload buttons on search page if user has no account with write permissions.

farmas:


src/app/search/search-page/search-page.component.spec.ts
```diff
@@ -149,6 +159,38 @@ describe('SearchPageComponent', () => {
...
+    testUser.accounts.set('TEST_ACCOUNT', 'r');
+    fixture = TestBed.createComponent(SearchPageComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+
+    uploadButton = fixture.debugElement.nativeElement.querySelectorAll('.cs-upload-new-document-button');
+    expect(uploadButton.length).toEqual(0);
+
+    // Verify button is displayed for write permission as object.
+    (<Spy>userServiceSpy.getUser).and.returnValue({ accounts: { TEST_ACCOUNT: 'rwd' } });
```
farmas, line 185: The reason for this test is that when accounts are loaded from backend the accounts is a simple object, NOT a Map<string, string>

src/app/search/search-page/search-page.component.ts
```diff
@@ -245,6 +249,24 @@ export class SearchPageComponent implements OnInit, OnDestroy, AfterViewInit {
...
+   * Returns true if user has at least one account with write permission.
+   * Note: possible permissions are "r", "rw", "rwd", and "admin".
+   */
+  private doesUserHasWritePermission() {
+    const user = this.userService.getUser();
+    const accounts = user && user.accounts;
+
+    if (accounts) {
+      // Test for Map or object dictionary.
+      const permissions = accounts.values ? Array.from(accounts.values()) : Object.keys(accounts).map((k) => accounts[k]);
```
farmas, line 262: This is why we shouldn't use Map or Set in public interfaces.

Back-end only return object dictionaries, but we cast the object into our interface that has Map. So at runtime, we "think" we have a Map, but we really have a regular Object.

jfshenuw (approved):

PR #1175, opened 2021-03-01
CAB-4228: Adds a warning message to the UUI header set via config.

farmas:
Summary:
This change follows the pattern described [here](https://angular.io/guide/component-interaction#parent-and-children-communicate-via-a-service) and adds a mechanism for components to communicate changes in state to the main app component. One of these states is the warning header message.

Notes:
- As always, the commit hook that runs 'prettier' re-formatted other lines in the files touched that are outside the scope of this change.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -1,11 +1,12 @@
 import { ChangeDetectorRef, Component } from '@angular/core';
 import { ProgressService } from './shared/providers/progress.service';
+import { ApplicationStateService } from './shared/providers/application-state.service';
 import { Angulartics2GoogleTagManager } from 'angulartics2/gtm';
 
 @Component({
   selector: 'app-root',
   templateUrl: './app.component.html',
-  styleUrls: ['./app.component.css']
+  styleUrls: ['./app.component.css'],
 })
 export class AppComponent {
   title = 'app';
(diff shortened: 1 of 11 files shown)


src/app/app.component.ts
```diff
@@ -14,22 +15,29 @@ export class AppComponent {
...
     });
-    progressService.value$.subscribe(value => {
+    progressService.value$.subscribe((value) => {
       this.value = value;
       this.changeDetector.detectChanges();
     });
+
+    appStateService.warningHeaderMessageChanged$.subscribe((message) => {
+      this.warningHeaderMessage = message;
+    });
```
farmas, line 41: This action could have been added to the existing 'progressService' that already acts a broker between application state and child components, but I thought that the service was too specific for progress operations that I decided to introduce a generic application state service.

jfshenuw:
> Summary:
> This change follows the pattern described [here](https://angular.io/guide/component-interaction#parent-and-children-communicate-via-a-service) and adds a mechanism for components to communicate changes in state to the main app component. One of these states is the warning header message.
> 
> Notes:
>

jfshenuw (commented):
Suppose, I see an outage message when I go to the UUI. Suppose the outage is resolved in 10 minutes. Would the outage message disappear automatically or would it disappear only after I refresh the UUI (to get the updated configuration from the backend)?

Would one need to update the UUI config for each of the 10+ profiles before and after an incident?

farmas:
(I don't know how to reply directly to your comment).

As discussed in CAB-4228, this iteration will support setting a separate message on each profile. If the same message is needed on all profiles, each would need to be edited separately.

As to the first question, the message does not disappear automatically, the user would need to refresh the site.

jfshenuw (approved):
The code change looked good to me. You may need to test the latency between the time when a change is uploaded to S3 and the time the change is visible on the UI, due to caching in the UI and in the backend (Profile API).

I fell that it is important to be able to upload a single file. Updating each profile during an incident is both time consuming and error-prone.

PR #1194, opened 2021-04-30
CAB-4324 view and download persisted files from wcc directly 

jfshenuw:
view and download persisted files from wcc directly  when wccUrl is set in the environment

### DIFF
--- a/src/app/content/content-view/content-view.component.html
+++ b/src/app/content/content-view/content-view.component.html
@@ -39,7 +39,7 @@
       <div *ngSwitchCase="'pdfUrl'">
 
         <!-- iframe is required to get file directly from WCC -->
-        <pdf-viewer *ngIf="!showPdfInIframe && !(getFileFromWcc && contentObject.persisted)"
+        <pdf-viewer *ngIf="!showInIframe()"
                     [src]="contentObject.url"
                     [(page)]="pageNumber"
                     [external-link-target]="'blank'"
@@ -56,7 +56,7 @@
                     class="cs-preview-max-height"
         ></pdf-viewer>
 
-        <div class="iframe-container" *ngIf="showPdfInIframe || (getFileFromWcc && contentObject.persisted)"  >
+        <div class="iframe-container" *ngIf="showInIframe()"  >
           <iframe [src]="getIframeUrlForPDF(useWccUrlIfNecessary(contentObject.url))" height="100%" name="iframe-content-display" scrolling="yes">Loading...</iframe>
         </div>
 
--- a/src/app/content/content-view/content-view.component.html
+++ b/src/app/content/content-view/content-view.component.html
@@ -57,7 +57,7 @@
         ></pdf-viewer>
 
         <div class="iframe-container" *ngIf="showPdfInIframe || (getFileFromWcc && contentObject.persisted)"  >
-          <iframe [src]="getIframeUrlForPDF(contentObject.url)" height="100%" name="iframe-content-display" scrolling="yes">Loading...</iframe>
+          <iframe [src]="getIframeUrlForPDF(useWccUrlIfNecessary(contentObject.url))" height="100%" name="iframe-content-display" scrolling="yes">Loading...</iframe>
         </div>
 
       </div>
--- a/src/app/content/content-view/content-view.component.html
+++ b/src/app/content/content-view/content-view.component.html
@@ -38,7 +38,8 @@
 
       <div *ngSwitchCase="'pdfUrl'">
 
-        <pdf-viewer *ngIf="!showPdfInIframe"
+        <!-- iframe is required to get file directly from WCC -->
+        <pdf-viewer *ngIf="!showPdfInIframe && !(getFileFromWcc && contentObject.persisted)"
                     [src]="contentObject.url"
                     [(page)]="pageNumber"
                     [external-link-target]="'blank'"
@@ -55,7 +56,7 @@
                     class="cs-preview-max-height"
         ></pdf-viewer>
 
-        <div class="iframe-container" *ngIf="showPdfInIframe"  >
+        <div class="iframe-container" *ngIf="showPdfInIframe || (getFileFromWcc && contentObject.persisted)"  >
           <iframe [src]="getIframeUrlForPDF(contentObject.url)" height="100%" name="iframe-content-display" scrolling="yes">Loading...</iframe>
         </div>
 
--- a/src/app/content/content-view/content-view.component.html
+++ b/src/app/content/content-view/content-view.component.html
@@ -39,7 +39,7 @@
       <div *ngSwitchCase="'pdfUrl'">
 
         <!-- iframe is required to get file directly from WCC -->
-        <pdf-viewer *ngIf="!showInIframe()"
+        <pdf-viewer *ngIf="!shouldShowInIframe()"
                     [src]="contentObject.url"
                     [(page)]="pageNumber"
                     [external-link-target]="'blank'"
@@ -56,7 +56,7 @@
                     class="cs-preview-max-height"
         ></pdf-viewer>
 
-        <div class="iframe-container" *ngIf="showInIframe()"  >
+        <div class="iframe-container" *ngIf="shouldShowInIframe()"  >
           <iframe [src]="getIframeUrlForPDF(useWccUrlIfNecessary(contentObject.url))" height="100%" name="iframe-content-display" scrolling="yes">Loading...</iframe>
         </div>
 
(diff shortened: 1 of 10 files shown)


src/app/content/edit-page/edit-page.component.html
```diff
@@ -44,10 +44,16 @@ <h2>{{pageConfig.pageName}}</h2>
...
           </div>
+          <!-- set getFileFromWcc to true here to enable this feature for all customers -->
+          <!-- we may use pageConfig later if we find it necessary to enable this feature for some customers only -->
           <div *ngIf="pageConfig?.viewPanel" fxFlex="auto">
             <div class="cs-content-view-wrapper">
               <div class="cs-content-view-wrapper-insert">
-                <app-content-view [contentObject]="contentObject" [allowPageByPageMode]="pageConfig?.allowPageByPageMode"></app-content-view>
+                <app-content-view
+                    [contentObject]="contentObject"
+                    [getFileFromWcc]="true"
```
agagne, line 54: Would it be better to add this parameter to the class ContentPageConfig rather than hard coding it within the html?

jfshenuw: The ContentPageConfig is shared by edit-page and create-page. As a result, the getFileFromWcc would default to false. To set this flag to true for edit-page only, I would still need to hard-code the value here, unless I add "getFileFromWcc: true" to all UUI configuration files. Once we decide to get file from WCC for some customers only, we would do the following: 1. add the field to ContentPageConfig, as you suggested. 2. use the value from ContentPageConfig instead. 3. Add the flag to some UUI conf files to turn them on for those customers. 

When I think about this again, it may be a good idea to turn this feature on for OAP only initially, instead of for all customers. What do you think?

src/app/content/content-view/content-view.component.html
```diff
@@ -38,7 +38,8 @@
 
       <div *ngSwitchCase="'pdfUrl'">
 
-        <pdf-viewer *ngIf="!showPdfInIframe"
+        <!-- iframe is required to get file directly from WCC -->
+        <pdf-viewer *ngIf="!showPdfInIframe && !(getFileFromWcc && contentObject.persisted)"
```
farmas, line 42: Consider abstracting this in a calculated property, since the condition seems to be inverted in the div below.

src/app/content/content-view/content-view.component.ts
```diff
@@ -153,10 +158,20 @@ export class ContentViewComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   getIframeUrlForPDF(url: any) {
```
farmas, line 160: If 'feels' like the intent of this method is simply to sanitize the passed in URL. Instead of adding the functionality of calculating the WCC url here, consider doing that work in a separate method and leave this for sanitize only. It could simplify the unit test since you wouldn't have to deal with the sanitizer.

src/app/content/content-view/content-view.component.spec.ts
```diff
@@ -135,6 +138,40 @@ describe('ContentViewComponent', () => {
     expect(webUrl).toBe('testUrl/456');
   });
 
+  it('should build wcc url', () => {
```
farmas, line 141: It is not clear what is the difference between 'building wcc url' and 'use wcc url' in the following test. Could you help clarify what is the intended use of each?

src/app/content/content-view/content-view.component.ts
```diff
@@ -153,10 +158,20 @@ export class ContentViewComponent implements OnInit, OnChanges, OnDestroy {
...
+    // can get file from wcc only it is already persisted to WCC
+    if (this.getFileFromWcc && this.contentObject && this.contentObject.persisted && this.contentObject.itemId) {
+      url = environment.wccUrl + this.wccGetFilePath + '&dDocName=' + this.contentObject.itemId + '&Rendition=Web&noSaveAs=1';
+    }
     return this.sanitizer.bypassSecurityTrustResourceUrl(url + '#view=FitH');
   }
 
   buildUrl({ itemId, webViewable = true, useOriginalFilename = true, disposition }: FileUrlParameters): string {
```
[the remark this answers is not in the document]
jfshenuw: buildUrl is used to set download url only and the disposition is always set to "attachment" by the caller. I could add another statement to add "noSaveAs=1" when disposition is not "attachment", if you think it is necessary.

farmas (dismissed):
Left some comments for your consideration. Up to you whether to address them.

src/app/content/content-view/content-view.component.ts
```diff
@@ -153,10 +158,20 @@ export class ContentViewComponent implements OnInit, OnChanges, OnDestroy {
...
+    if (this.getFileFromWcc && this.contentObject && this.contentObject.persisted && itemId) {
+      let url = environment.wccUrl + this.wccGetFilePath + '&dDocName=' + itemId;
```
agagne, line 170: - Should this be moved into the `contentService`, similar to the `contentService.getFileUrl` below

src/app/content/content-view/content-view.component.ts
```diff
@@ -153,10 +158,20 @@ export class ContentViewComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   getIframeUrlForPDF(url: any) {
+    // can get file from wcc only it is already persisted to WCC
+    if (this.getFileFromWcc && this.contentObject && this.contentObject.persisted && this.contentObject.itemId) {
```
agagne, line 162: Is there a reason we are modifying the url at this point rather than when it is being set on the contentObject?

jfshenuw: See my reply to your earlier question.

jfshenuw: I set the URL on contentObject initially and it broke the file replacement.

src/app/content/content-view/content-view.component.ts
agagne: @jfshenuw I think centralizing the url in a service makes a lot of sense, we do it for the getFileUrl.. Why would you want the CreatePage to handle this logic differently?

jfshenuw: getting file from wcc is valid only for viewing/downloading file already in wcc. It is not valid for CreatePage. It is not valid for file replacement, where the file is from local file system.

src/app/content/edit-page/edit-page.component.html
jfshenuw: added getFileFromWcc flag to ContentPageConfig and used pageConfig flag in the html.

src/app/content/content-view/content-view.component.html
jfshenuw: moved the logic to a function in .ts.

src/app/content/edit-page/edit-page.component.html
```diff
@@ -44,10 +44,16 @@ <h2>{{pageConfig.pageName}}</h2>
             </div>
 
           </div>
+          <!-- set getFileFromWcc to true here to enable this feature for all customers -->
+          <!-- we may use pageConfig later if we find it necessary to enable this feature for some customers only -->
```
farmas, line 48: Looks like we are using PageConfig already, sounds like these comments no longer apply.

src/app/content/content-view/content-view.component.ts
```diff
@@ -152,11 +157,31 @@ export class ContentViewComponent implements OnInit, OnChanges, OnDestroy {
...
+      url = environment.wccUrl + this.wccGetFilePath + '&dDocName=' + this.contentObject.itemId + '&Rendition=Web&noSaveAs=1';
+    }
+
+    return url;
+  }
+
+  // Iframe is required to get file from WCC
+  showInIframe() {
```
farmas, line 170: "showInInIframe()" implies that this is an action method. Consider renaming to "canShowInIframe" or something similar. Also could be a getter function.

src/app/content/content-view/content-view.component.ts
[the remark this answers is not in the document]
farmas: If this is for download only, why is the function taking a 'disposition' argument and passing it to the contentService.getFileUrl() method?

farmas (dismissed):
Left you comments to consider.

src/app/content/content-view/content-view.component.ts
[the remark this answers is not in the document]
jfshenuw: The argument could be removed from the 2 calls to buildUrl and then hard-code 'attachment' when calling getFileUrl. This would be a small refactoring work. Since UUI is going away, I don't know it is worthwhile.

src/app/content/content-view/content-view.component.ts
jfshenuw: "canShowInIframe" indicates it is optional, not mandatory. I considered using "needToShowInIframe" but ended with the shorter "showInIframe". Making it a getter function is a good idea.

jfshenuw: I renamed it to shouldShowInIframe.

src/app/content/edit-page/edit-page.component.html
jfshenuw: removed outdated comments.

farmas (approved):

PR #1195, opened 2021-05-03
CAB-4325: Restricts input to number and adds $ prefix for displayType=currency fields.

farmas:
input type="number" rejects non-numerical entries. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number.

A "$" prefix is also added to make it clear to user that the field represents a currency amount. See screen shot below.


farmas:
Screen shot that shows the rendering of the field:

Screen Shot 2021-05-03 at 3 32 25 PM

agagne:
> 
> 
> Screen shot that shows the rendering of the field:
> 
> Screen Shot 2021-05-03 at 3 32 25 PM

Thanks for adding the screenshot! We should adopt this practice for future PRs.

agagne (approved):
Looks good

PR #1196, opened 2021-05-04
CAB-4325: Adds regex validator for number fields.

farmas:
Testing revealed that input with type="number" doesn't work well in all browsers. Switched the solution to use a regex validator instead. See screen shot below for experience.

### DIFF
--- a/src/app/content/content-metadata/content-metadata.component.ts
+++ b/src/app/content/content-metadata/content-metadata.component.ts
@@ -22,6 +22,8 @@ const nonEmptyFormValidator: ValidatorFn = (theForm: FormGroup): ValidationError
   return null;
 };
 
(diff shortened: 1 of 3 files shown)


agagne (approved):
good idea

farmas:
Screen Shot 2021-05-04 at 10 50 29 AM

src/app/content/content-metadata/content-metadata.component.ts
```diff
@@ -22,6 +22,8 @@ const nonEmptyFormValidator: ValidatorFn = (theForm: FormGroup): ValidationError
   return null;
 };
 
+const NUMBER_REGEX_PATTERN = '^[0-9]+(\\.[0-9]+)?$';
```
farmas, line 25: Could be made fancier, but I thought this was a balance of pattern simplicity while still being correct.

agagne (approved):

